PR #2, opened 2015-02-05
Capitalize patchmanager entry on the jolla-settings menu

NickCis:
The commit only changes "patchmanager" to "Patchmanager" in order to have the entry capitalized in the jolla-settings app. (All entries are with capital letter, i found it ugly to have patchmanager entry starting with lowercase").

### DIFF
--- a/src/plugin/patchmanager.json
+++ b/src/plugin/patchmanager.json
@@ -4,7 +4,7 @@
         {
             "path": "system_settings/look_and_feel/patchmanager",
             "type": "page",
-            "title": "patchmanager",
+            "title": "Patchmanager",
             "translation_id": "settings-patchmanager-patchmanager",
             "icon": "/usr/share/patchmanager/icons/icon-m-patchmanager.png",
             "order": 2000,


SfietKonstantin:
Long discussion. The app is called patchmanager, just like a command line app (cd, ls, top etc.) Calling it Patchmanager is a bit weired IMO.

NickCis:
I understand what you are saying, but, to the coherent with all the jolla-settings, it would be nice to have the list entry capitalized.
A similar case is the one of sailfish-utilities ( https://github.com/sailfishos/sailfish-utilities ), the app is called sailfish-utilities, but the entry in the jolla-settings app is "Utilities" ( https://github.com/sailfishos/sailfish-utilities/blob/master/utilities.json.in ).
I think that as all entries in the list are capitilized it would be nice, in order to be cohesive with all the app aesthetic, to have this one capitalized too.

CODeRUS:
LGTM // Menu entry name should be capitalized

PR #4, opened 2015-03-11
don't repeat yourself

benaryorg:


src/qml/patchmanager.cpp
```diff
@@ -115,9 +115,7 @@ void PatchManager::restartServices()
     }
 
     if (m_homescreenNeedRestart) {
-        QStringList arguments;
-        arguments << "--user" << "restart" << "lipstick.service";
-        QProcess::startDetached("systemctl", arguments);
+        restartLipstick();
```
SfietKonstantin: You can also remove the second if (m_homescreenNeedRestart) that is following this if

SfietKonstantin:
Interesting error ... Only one comment and LGTM

src/qml/patchmanager.cpp
CODeRUS: I'am voting for rewriting it to use D-Bus calls instead of QProcess.

SfietKonstantin: Is there a DBus call to restart lipstick ? Or do you want to go with a call to systemd ?

CODeRUS: call to systemd. i'll look into this later today.

SfietKonstantin: Thanks, waiting for PR

benaryorg:
Just ping me when it's time to close this PR.…

CODeRUS:
you can close it. but today i have some weird problems with SDK.

CODeRUS:
Closed because of https://github.com/sailfishos-patches/patchmanager/pull/8

PR #5, opened 2016-08-12
Added unapplyall patch and dbus method to daemon

CODeRUS:
@SfietKonstantin

### DIFF
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -112,6 +112,8 @@ dbus-send --system --type=method_call \
 %{_datadir}/%{name}/tools
 %{_datadir}/dbus-1/
 %{_sysconfdir}/dbus-1/system.d/
+%{_datadir}/patchmanager/patches/sailfishos-patchmanager-unapplyall/patch.json
+%{_datadir}/patchmanager/patches/sailfishos-patchmanager-unapplyall/unified_diff.patch
 # >> files
 # << files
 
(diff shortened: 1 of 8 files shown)


rpm/patchmanager.spec
```diff
@@ -61,6 +61,17 @@ on your SailfishOS device easily. This package
 contains the Jolla settings plugin.
 
 
+%package -n sailfishos-%{name}-unapplyall
```
SfietKonstantin: Make it part of patchmanager. Do not create a new package

src/unapply/unified_diff.patch
```diff
@@ -0,0 +1,40 @@
...
+         contentHeight: content.height
+@@ -250,10 +267,7 @@ Page {
+                             }
+                         }
+                     } else if (storeIf.mayInstall) {
+-                        // Let the UI smoothly fade out first before doing anything scary
+-                        shutdownBus.call("setShutdownMode", ["upgrade"])
+-                        // Clear the cache and the system update notification
+-                        storeIf.clearOsUpdateCheckCache()
++                        patchmanagerBus.call("unapplyAllPatches", [], unapplyAllCallback)
```
SfietKonstantin: Maybe there could be a problem here: what do we do when we failed to unapply patches ?

CODeRUS: we can't upgrade sfos until some patch failed to unapply, that's the point.

rpm/patchmanager.spec
CODeRUS: hm, okay

src/unapply/unified_diff.patch
SfietKonstantin: My fault, I didn't reviewed it correctly ...

CODeRUS:
done fixing comments :)

PR #8, opened 2016-08-20
Added Unapply all patches to UI

CODeRUS:
and did small refactoring in code

@SfietKonstantin


rpm/patchmanager.yaml
```diff
@@ -32,24 +32,24 @@ Files:
 - '%{_datadir}/patchmanager/patches/sailfishos-patchmanager-unapplyall/patch.json'
 - '%{_datadir}/patchmanager/patches/sailfishos-patchmanager-unapplyall/unified_diff.patch'
 SubPackages:
-- Name: ui
```
SfietKonstantin: Why :(

CODeRUS: Because nobody using app icon, settings applet is better :)

SfietKonstantin:
Need a bit more time to review. This PR is rather large

Issue #35, opened 2018-11-18
[PM3 simplification suggestion] Remove ability to manually reorder Patches


Olf0:
**Prelude:**
When Patchmanager 3 was originally released, IIRC it did "exact patching" (in contrast to PM2).  Hence the ability to determine the order, in which Patches are applied, was needed.  This was nicely implemented by being able to manually drag Patches in Patchmanager's list view (primary window) into their desired position.

Meanwhile PM3 reverted to use "fuzzy patching" months ago (IIRC on Ancelad's request at TMO), hence the sequence in which the Patches are applied does not matter anymore (as it is the case for PM2).  At least when experimenting with 50+ Patches installed under SFOS 2.2.0, 2.2.1 and 3.0.0 on a Jolla 1 phone and an Xperia X, the order of the Patches never made any difference (and why should it, as it is not an issue with PM2?).

**Advantages:**
* While manually ordering Patches by dragging them is cool, it complicates the GUI for inexperienced users.
* The extra code for dragging Patches may have bugs (just as any code).
* The extra code for dragging Patches poses a permanent maintenance burden (just as any code).
* Without a fixed order of the Patches (plus nicely grouped in their categories in PM2), the list of installed Patches is almost incomparable between devices at the GUI.

**Suggestions:**
Please revert the list of Patches (in Patchmanager's primary window) to a fixed order, as in Patchmanager 2: Alphabetically sorted and grouped in alphabetically sorted categories.  PM2 is doing that fine and looking good in this regard.
Also eliminate the ability to manually reorder Patches by dragging them at the GUI in PM's list view.

Furthermore it would be nice to have the category of a specific Patch also displayed in an additional section "Category" (to the extant ones: Name, Version, Maintainer, Description, Patch log) in its Patch details sub-page.

PR #44, opened 2020-01-04
Sort web patches by category and name

mentaljam:
Sort patches by localized category and displayed name. The "other" category is always displayed at the end of the list.

### DIFF
--- a/src/qml/WebCatalogPage.qml
+++ b/src/qml/WebCatalogPage.qml
@@ -41,6 +41,7 @@ Page {
     property var versions
     property string search
     property bool searchVisible
+    property bool sortByDate: true
 
     onStatusChanged: {
         if (status == PageStatus.Active) {
@@ -84,13 +85,21 @@ Page {
                     searchVisible = !searchVisible
                 }
             }
+            MenuItem {
+                text: sortByDate ? qsTranslate("", "Sort by Category") : qsTranslate("", "Sort by Date")
+                onClicked: {
+                    sortByDate = !sortByDate
+                }
+            }
+
         }
 
         header: Component {
             Column {
                 width: view.width
                 PageHeader {
-                    title: container.author ? qsTranslate("", "%1 patches").arg(container.author) : qsTranslate("", "Web catalog")
+                    title: container.author ? qsTranslate("", "%1 patches").arg(container.author) : qsTranslate("", "Web catalog") 
+                    description: container.sortByDate ? qsTranslate("", "(by date updated)") : qsTranslate("", "(by category)")
                 }
 
                 SearchField {
@@ -130,15 +139,22 @@ Page {
         }
         model: WebPatchesModel {
             id: patchModel
-            queryParams:  container.author ? { 'author': container.author }
-                                           : container.search ? { 'display_name__contains': container.search }
-                                                              : {}
+            sorted: !container.sortByDate
+            queryParams: {
+                var p = {};
+                if (container.author)  p = { 'author': container.author };
+                if (container.search)  p = { 'display_name__contains': container.search };
+                return p;
+            }
         }
         section.criteria: ViewSection.FullString
         section.delegate: SectionHeader {
-            text: qsTranslate("Sections", section)
+            text: container.sortByDate ? "" : qsTranslate("Sections", section)
+            font.capitalization: Font.Capitalize
+            height: text.length > 0 ? implicitHeight : 0
+            visible: text.length > 0
         }
-        section.property: "category"
+        section.property: container.sortByDate ? "undefined" : "category"
         currentIndex: -1
         spacing: Theme.paddingSmall
 
@@ -167,7 +183,7 @@ Page {
                     width: parent.width
                     Label {
                         id: nameLabel
-                        width: parent.width - authorLabel.width - Theme.paddingMedium
+                        width: parent.width - secondaryLabel.width - Theme.paddingMedium
                         text: model.display_name
                         color: background.down ? Theme.highlightColor : Theme.primaryColor
                         font.bold: isInstalled
@@ -175,11 +191,11 @@ Page {
                     }
 
                     Label {
-                        id: authorLabel
+                        id: secondaryLabel
                         anchors.right: parent.right
                         anchors.verticalCenter: parent.verticalCenter
                         font.pixelSize: Theme.fontSizeExtraSmall
-                        text: model.author
+                        text: container.sortByDate ? model.category : model.author
                         color: Theme.secondaryHighlightColor
                     }
                 }
--- a/src/qml/webpatchesmodel.cpp
+++ b/src/qml/webpatchesmodel.cpp
@@ -55,6 +55,7 @@ WebPatchesModel::WebPatchesModel(QObject * parent)
     }
 
     _completed = false;
+    _sorted = false;
     _nam = new QNetworkAccessManager(this);
 }
 
@@ -73,6 +74,15 @@ void WebPatchesModel::setQueryParams(const QVariantMap & queryParams)
         _queryParams = queryParams;
         emit queryParamsChanged(_queryParams);
 
+        if (_completed) {
+            componentComplete();
+        }
+    }
+}
+void WebPatchesModel::setSorted(const bool & sorted) {
+    if (sorted != _sorted) {
+        _sorted = sorted;
+        emit WebPatchesModel::sortedChanged();
         if (_completed) {
             componentComplete();
         }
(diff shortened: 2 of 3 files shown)


nephros:
Just tested this against current master, appears to me working nicely.

It would be nice if the sorting could be toggled though, so the 'sort by modification date' view is available as well, because that has its uses.

And while we're at it, maybe 'sort by author' too?

Olf0:
Just to put these details and ideas into perspective:
- This PR solves an annoying issue, that it becomes really hard to find a specific patch, if one has many installed. 
- > It would be nice if the sorting could be toggled though, so the 'sort by modification date' view is available as well, because that has its uses.
  
  Yes, that would be "nice" and may be useful for special requirements. 
- > And while we're at it, maybe 'sort by author' too?
  
  This would be a feature, "because we can".  I cannot see a really practical application of this, but "why not".  But I am sure it is not important at all.

nephros:
> Just to put these details and ideas into perspective:
> 
> * This PR solves an annoying issue, that it becomes really hard to find a specific patch, if one has many installed.

Precisely.
And as we cannot know which criteria a user might want to apply to locate a specific patch, we can give them options, instead of presuming there's only one use case.

> * > It would be nice if the sorting could be toggled though, so the 'sort by modification date' view is available as well, because that has its uses.
>   
>   
>   Yes, that would be "nice" and may be useful for special requirements.

We have a published/established implementation of the view (sorted by modification date, or rather, I think, whatever order the web service delivers), it may not be ideal, but it is what users know now.
It is also a useful view when the  "you have updates" notification happens.

We also have several properties in the model which lend themselves to sorting (others, not so much).

> * > And while we're at it, maybe 'sort by author' too?
>   
>   
>   This would be a feature, "because we can".  I cannot see a really practical application of this, but "why not".  But I am sure it is not important at all.

It has just as much a justification as sort by category. Users may want to locate patches categorized by either, or other properties.

nephros:
For your consideration:

https://github.com/nephros/patchmanager/tree/pr-44

mentaljam:
Hello everyone!

Sorry, I cannot participate in this PR for now. By I totally agree that there should be an option to toggle sorting with sort by date by default. The only thing I can add, I think the category sections should be enabled only when sorting by category.

b100dian:
I also agree that 'sort by modification date' should continue to be available. @mentaljam thanks for your PR - @nephros do you want to open one with your settings and close this? I've tested it and I have two comments, one nit about code formatting and the other one about squeezing the secondaryLabel:) Also, I would started adding a storage for the property (e.g. catalogSort:none or category) for further expansion but am still pondering if the settings are only to be changed through dbus as today, hmm

Olf0:
> > > It would be nice if the sorting could be toggled though, so the 'sort by modification date' view is available as well, because that has its uses.
> >   
> >   Yes, that would be "nice" and may be useful for special requirements.
> 
> We have a published/established implementation of the view (sorted by modification date, or rather, I think, whatever order the web service delivers), it may not be ideal, but it is what users know now.
> It is also a useful view when the "you have updates" notification happens.

Well, I never understood the order of the patches in the we-catalog: It always appeared to me as "without any proper order" (or as you put it "whatever order the web service delivers"), but I can tell that this order is static (i.e., not "by modification date"; I can also tell that "by creation date in the web-catalog" is unlikely, because freshly submitted (new) patches did appear anywhere on that list).
But I can follow your rationale, especially because that order is somewhat static (except for new patches, which may appear anywhere), so users are used to that unordered order.

> We also have several properties in the model which lend themselves to sorting (others, not so much).
> 
> > > And while we're at it, maybe 'sort by author' too?
> >     
> >   This would be a feature, "because we can".  I cannot see a really practical application of this, but "why not".  But I am sure it is not important at all.
> 
> It has just as much a justification as sort by category. Users may want to locate patches categorized by either, or other properties.

Yes, that is why I tried to point out potential priorities / "put things into perspective": There are quite some properties which may be used for ordering the list of patches.
This just addresses the patches from the web-catalog, but IMO reworking this should be a part of a concept, how patches are ordered; not only in the web-catalog's list, but also at the list of installed patches.
But the list of installed patches also affects the order in which patches are applied; see also the related issue #35.  But almost no user is aware of that, because it is not documented.  Plus I doubt, that this functionality is really still necessary (that is what issue #35 is about).

Look, I really do not want to get in an argument here: Implement whatever sorting options for the web-catalog's list you like, that cannot do any harm.
But I rather would like to discuss a concept which sorting options are crucial and how they are applied to the web-catalog's list and the list of installed patches alike (i.e., offering the same sorting options for both, plus maybe "installed list" specific ones as e.g., "from RPM vs. web-catalog" or "order in which they are applied" (if that is really still necessary, technically)), because I know very well that managing > 30 patches is not nice at all (especially finding a specific patch or group of patches), currently.

nephros:
Thanks all for the copious input.

I'll not yet make a PR of it, I'd like to flesh it out a bit more.

I'm thinking of exposing an enum to QML to select the sorting mode instead of just a toggle, and adding a couple of more sorters to implement @Olf0's suggestions.

OTOH, we could also add what we have now, and do the improved version later. Thoughts?

On the topic of sorting tha main patch view, here we are blocking indeed on #35, and we need to decide which way to go here.
A hybrid approach is imaginable (keeping the patch order internally, while displaying a different, sorted order in the UI), but that seems unnecessarily confusing for both users and implementors.

mentaljam:
> A hybrid approach is imaginable (keeping the patch order internally, while displaying a different, sorted order in the UI), but that seems unnecessarily confusing for both users and implementors

IMHO Implementing [QSortFilterProxyModel](https://doc.qt.io/archives/qt-5.6/qsortfilterproxymodel.html) with options (an enum as @nephros mentioned) is not so complicated

nephros:
> I also agree that 'sort by modification date' should continue to be available. @mentaljam thanks for your PR - @nephros do you want to open one with your settings and close this? I've tested it and I have two comments, one nit about code formatting and the other one about squeezing the secondaryLabel:) Also, I would started adding a storage for the property (e.g. catalogSort:none or category) for further expansion but am still pondering if the settings are only to be changed through dbus as today, hmm

Okay, force-pushed a version rebased to current master.

@b100dian please review with your nits.

Working on the improved version meanwhile, but that will take some time.

src/qml/webpatchesmodel.cpp
```diff
@@ -94,9 +116,43 @@ void WebPatchesModel::componentComplete()
...
+                    const auto amap = a.toMap();
+                    const auto bmap = b.toMap();
+
+                    const auto acat = amap[category].toByteArray();
+                    const auto bcat = bmap[category].toByteArray();
+
+                    if (acat == bcat)
+                    {
+                    // If categories are equal then sort by name
+                    return compareStrings(amap[name].toString(), bmap[name].toString());
```
b100dian, line 138: so the nit about formatting is:

In this lambda the tabbing of the code inside the curly braces does not look like the rest of the file (or project - at least on cursory view). Also on line 148 it doesn't look like the rest of the function.

I do not want to start a 'code conventions' discussion here, let's go to 'ambient conventions' for the moment, what do you think?

src/qml/WebCatalogPage.qml
```diff
@@ -167,19 +183,19 @@ Page {
...
                     }
 
                     Label {
-                        id: authorLabel
+                        id: secondaryLabel
                         anchors.right: parent.right
                         anchors.verticalCenter: parent.verticalCenter
                         font.pixelSize: Theme.fontSizeExtraSmall
-                        text: model.author
+                        text: container.sortByDate ? model.category : model.author
```
b100dian, line 198: The question here is how do you see this in the future:
Right now, you have the equivalent of:
 - no sorting - you see sections == categories, secondary label == author
 - by category - you don't see sections, secondary label = category

How does this 'scale'? What if you introduce
 - by author - you don't see sectons, secondary label = category (?)

b100dian (commented):
Awesome that was pushed to the same PR.

I'm favorable to approaching this incrementally, we are not 'at rush' anymore but we should not start gold-plating either.

The way this PR is today has the advantage of 1. adding a needed feature and 2. preserving the original behavior for comparison / further testing.

Adding a last-modified order would not hurt. Adding more than a boolean for the sort order would take more time. Adding all of the options Olf outlined above even more. Linking this to the List of Installed patches - again, even more.

I have sufficient bandwidth for testing and reviewing, but I cannot say the same for coding at the moment - so I let you @nephros  draw the line where to stop for this PR. Thanks!

src/qml/WebCatalogPage.qml
nephros: My thinking was doing away with the boolen completely, introducing an enum or a State in the QML, and using that to steer sorting.

Sorting itself should also be done in the model so it doesn't trigger a new request to the server if changed - as @mentaljam suggested earlier.

But that is something to tackle in the future when an improved version takes shape.

nephros:
> Awesome that this was pushed to the same PR.
> 
> I'm favorable to approaching this incrementally, we are not 'at rush' anymore but we should not start gold-plating either.
> 
> The way this PR is today has the advantage of 1. adding a needed feature and 2. preserving the original behavior for comparison / further testing.
> 
> Adding a last-modified order would not hurt. Adding more than a boolean for the sort order would take more time. Adding all of the options Olf outlined above even more. Linking this to the List of Installed patches - again, even more.
> 
> I have sufficient bandwidth for testing and reviewing, but I cannot say the same for coding at the moment - so I let you @nephros draw the line where to stop for this PR. Thanks!

Alright, my initial tries with doing the gold-plated version do not look promising.
So therefore I'll do a little cleanup on this as it stands and request to merge it as-is now.

b100dian (approved):
+1

PR #46, opened 2020-10-15
Adjust spelling and grammar a bit

Olf0:
A try to enhance the language of some dialogues and console output.


Olf0:
Please note and check (**before** merging):
1. I am not experienced in handling TS files.  So I did not know, if the statement `<TS language="fo_BA" version="2.1">` denotes the version of the translation file proper or some markup language version these files adhere to.  Hence I retained them as they all in their recent state, all at *version 2.1*.
If they have to be increased to *version 2.2*, please do so or let me know and I will.
2. I think I accidentally appened a carriage return (CR) at the end of each TS file.
If they have to be deleted, please let me know and I will remove them.
3. I am not completely happy with the "all" I inserted.  I think the "all"-aspect is important, but am now considering "all activated" (because "... applying all applied ..." would sound strange, almost like a circular conclusion or a self-fulfilling prophecy) to be more specific.
What do you think?

nephros:
The changes to the source files are OK, however AFAIK the workflow is something like

   source file(s) --> `lupdate src/qml/*.qml -ts src/qml/settings-patchmanager-en.ts` --> ts file gets picked up by transifiex --> translations happen on transifex --> `translations/pull.sh` can be used to get the translated parts back into the repo.

Correct?

(And I only know Weblate, but there it is also possible to pull from a git repo where the translated files are managed. I assume transifex has something similar.)

If that is so, corrections to the files should be done on transifex, not in the repo directly.

Or rather, just `src/qml/settings-patchmanager-en.ts` must be updated and the others will be later by the fine people translating.

Olf0:
> The changes to the source files are OK, 

Good.
But ...

> however AFAIK the workflow is something like 
> source file(s) --> `lupdate *.qml -ts masterlanguagefile.ts` --> ts file gets picked up by transifiex --> translations happen on transifex --> `translations/pull.sh` can be used to get the translated parts back into the repo.
> Correct?

Likely, because it sound reasonable and similar to the fragmets of Transifex-documentation I read.  But I only *know* Transifex from its web-frontend as a translator.

> (And I only know Weblate, but there it is also possible to pull from a git repo where the translated files are managed. I assume transifex has something similar.)
> 
> If that is so, corrections to the files should be done on transifex, not in the repo directly.

O.K., but unnecessarily updating them does no harm, because they will be overwritten, correct?

> Or rather, just `masterlanguagefile.ts` must be updated and the others will be later by the fine people translating.
> 
> EDIT: By the way, why are there `src/qml/settings-patchmanager.ts` _and_ `translations/settings-patchmanager.ts` with not quite but almost the same content? The former seems a little misplaced, but could be a master file, and the latter also has no translated content, so could also be a master file. The Transifex project clearly shows it must be a file of that name.

Looking at their history [`src/qml/settings-patchmanager.ts`](https://github.com/sailfishos-patches/patchmanager/commits/master/src/qml/settings-patchmanager.ts) seems to be the original file, while [`translations/settings-patchmanager.ts`](https://github.com/sailfishos-patches/patchmanager/commits/master/translations/settings-patchmanager-en.ts) is fetched from Transifex together with all other files in `translations`.
We will have to try this process in practice, if we gain administrative access at Transifex.

Olf0:
> > (And I only know Weblate, but there it is also possible to pull from a git repo where the translated files are managed. I assume transifex has something similar.)
> > If that is so, corrections to the files should be done on transifex, not in the repo directly.

This would mea to perform the enhancements in the target TS-files manually at Transifex: IMO a no-go, becausue it is many 10!
Also because many of these go hand in hand with the altered source strings.

b100dian:
> > > (And I only know Weblate, but there it is also possible to pull from a git repo where the translated files are managed. I assume transifex has something similar.)
> > > If that is so, corrections to the files should be done on transifex, not in the repo directly.
> 
> This would mean to perform the enhancements in the target TS-files manually at Transifex: IMO a no-go, because it is many 10! Also because many of these go hand in hand with the altered source strings.

IMO you shouldn't do the corrections in translated files - it would mean those old translations are for the new phrases (which are not).

nephros:
> > > > (And I only know Weblate, but there it is also possible to pull from a git repo where the translated files are managed. I assume transifex has something similar.)
> > > > If that is so, corrections to the files should be done on transifex, not in the repo directly.
> > 
> > 
> > This would mean to perform the enhancements in the target TS-files manually at Transifex: IMO a no-go, because it is many 10! Also because many of these go hand in hand with the altered source strings.
> 
> IMO you shouldn't do the corrections in translated files - it would mean those old translations are for the new phrases (which are not).

The question is, is there a way to get changes to -lang.ts files made locally picked up by Transifex?

Because the changes in this PR can be coaxed to merge to a current-day tree.
Also, I can do e.g. German translations, and would do them together with code changes.

But it's pointless to do if they will get lost by Transifex updates later.

Olf0:
> IMO you shouldn't do the corrections in translated files - 

As  a matter of fact, I did 3 years ago.

> it would mean those old translations are for the new phrases (which are not).

???
They are mostly fresh translations of the freshly enhanced source strings (plus maybe a few enhanced translations, where the source string was not altered).
Both, the source strings and the target strings did exist before.

My question is: How do we get that applied, before we alter the translation files (source & targets) massively?

b100dian:
I mean 
> Patchmanager appliquera automatiquement les patches dans 10 sec.

is not a translation for
>  Patchmanager will automatically apply all patches in 10 seconds.

("all" was added, but the French doesn't have it)

Olf0:
> My question is: How do we get that applied, before we alter the translation files (source & targets) massively?

O.K., looking at it, it is not that many strings.  I think, I can handle that manually, if you (@b100dian) assign me to *all* languages for Patchmanager at Transifex.
a. Is that feasible?

b. Do I understand correctly, that I shall delete all files, except for <dialog.qml> in this MR here?

If I receive a "Yes" for a *and* b, I will proceed.

b100dian:
My intuition says dialog.qml _AND_ translations/settings-patchmanager.ts are needed -> the latter is configured as "Source" file

nephros:
I suggest also picking up https://github.com/nephros/patchmanager/commit/66cef3119932139809efc6bb602bee550c16d56b, it makes `lupdate -pro patchmanager.pro` work correctly.

b100dian:
> The question is, is there a way to get changes to -lang.ts files made locally picked up by Transifex?

Let me get back to you with updates from my transifex for the b100dian fork

Olf0:
> My intuition says dialog.qml _AND_ translations/settings-patchmanager.ts are needed -> the latter is configured as "Source" file (i.e. source language)

This is crazy: It is supposed to *ease* the handling of translations, instead of copying the source strings manually from one file to another, before they can be translated via Transifex!

I gave it a new start in PR #76, but only altering <dialog.qml> for now.

b100dian:
> > The question is, is there a way to get changes to -lang.ts files made locally picked up by Transifex?
> 
> Let me get back to you with updates from my transifex for the b100dian fork

https://www.transifex.com/b100dian/patchmanager3-1/translate/#de/translations-settings-patchmanager-ts--master/334885475?q=text%3Aincompatible

 "Allow incompatible patches" appears with translation "Nicht kompatible Patches erlauben" but not reviewed. I think this is good.

> This is crazy: It is supposed to ease the handling of translations, instead of copying the source strings manually from one file to another

You just need to modify source files, but run a build (or manually `lupdate -pro patchmanager.pro` to generate the rest of the changes.
Of course this is not mandatory for all PRs, from time to time we need to update the strings though.

nephros:
> This is crazy: It is supposed to _ease_ the handling of translations, instead of copying the source strings manually from one file to another, before they can be translated via Transifex! angry
> 
> What for are the translateable strings marked with a `qsTranslate()` in <dialog.qml>?!? roll_eyes

Well, it's supposed to be the neutral exchange format, this way 

a) you don't have to hand over source code to a third party (which might not only not be preferred but depending on ownership, not possible)
b) .ts is a relatively simple XML format, can be handled by many tools, from Transifex to vi
c) those tools do not need to (re-)implement qt/qml parsing, that's done by `lupdate`, once

Olf0:
> > > The question is, is there a way to get changes to -lang.ts files made locally picked up by Transifex?
> > 
> > 
> > Let me get back to you with updates from my transifex for the b100dian fork
> 
> https://www.transifex.com/b100dian/patchmanager3-1/translate/#de/translations-settings-patchmanager-ts--master/334885475?q=text%3Aincompatible

This link requires me to log in Transifex (which is fine) and then displays a "404 - Not found". :disappointed: But I am successfully logged in at Transifex after that.
Maybe the link became broken by Copy&Paste, maybe it is incorrect, maybe I (simply "olf" at Transifex) lack the access rights, ...?

> "Allow incompatible patches" appears with translation "Nicht kompatible Patches erlauben" but not reviewed. I think this is good.

Why is that good?
It is not a change I provided: https://github.com/sailfishos-patches/patchmanager/pull/46/files#diff-195dfa71c9b8f1612e300365b42bad3446430de7d13c9bd0a824681be1e24849
> You just need to modify source files, 

Please define "source files" properly, see [questions 1 - 4 here](https://github.com/sailfishos-patches/patchmanager/pull/76#issuecomment-933012347)!

b100dian:
> This link requires me to log in Transifex (which is fine) and then displays a "404 - Not found". disappointed But I am successfully logged in at Transifex after that.
> Maybe the link became broken by Copy&Paste, maybe it is incorrect, maybe I (simply "olf" at Transifex) lack the necessary rights for acessing that, ...?

Sorry, new to transifex too, i thought this was open. Remember, this is a pet project where I'm learning. So I've added you as Reviewer to it, can you access the link now?

Olf0:
> Sorry, new to transifex too, i thought this was open. Remember, this is a pet project where I'm learning. So I've added you as Reviewer to it, can you access the link now?

Yes, thanks.

What becomes much more easily obviouis at Transifex (because they provide a nicer list), is that we should consolidate languages.  For the few strings Patchmanager has, nationalisms do not seem to be relevant:
- Just use `nl`, delete `nl_BE` (both seem to be the same level).
- Just use `fi`, delete `fi-FI` (which has less strings translated).
- Copy `fr_FR` over to `fr` (with no strings, currently) and delete `fr_FR`.
- Move `pt_BR` to `pt` (which does not exist, currently).
- Just use `sl`, delete `sl_SL` (both seem to be the same level).

=> 5 less languages with no loss.

Is that best done in Transifex (have not found a way) or in this repo (by deleting or renaming the corresponding TS-files)?

b100dian:
Noted. Let's see a first transifex update working first:p

Olf0:
O.K., so I will pose a MR the next couple of days, which introduces the changes from MR #76 to `translations/settings-patchmanager.ts`.

Is it O.K. to delete `src/qml/settings-patchmanager.ts` while at it?

Olf0:
> > This is crazy: It is supposed to _ease_ the handling of translations, instead of copying the source strings manually from one file to another, before they can be translated via Transifex! angry
> > What for are the translateable strings marked with a `qsTranslate()` in <dialog.qml>?!? roll_eyes
> 
> Well, it's supposed to be the neutral exchange format, this way
> 
> a) you don't have to hand over source code to a third party (which might not only not be preferred but depending on ownership, not possible) b) .ts is a relatively simple XML format, can be handled by many tools, from Transifex to vi c) those tools do not need to (re-)implement qt/qml parsing, that's done by `lupdate`, once

Thank you for your serious answer to my rather emotional outburst.
The takeaway for me is: I did not expect the task, which Transifex performs to be technically that simple.
While they provide a nice web-frontend, it is a proprietary service, which locks in the data and requires all its users to sign off all rights (i.e., Atlassian can do with the strings and the recorded user activities what they like).  But then, it is a big company selling web-services, this service seems to be commercially successful, hence they do maintain and run it properly.  The only point I do not trust Atlassian at, is to keep the "terms of use" the same: They love to press as much money out of their customers as possible.
@nephros, didn't you say, you have experience with pootle?  Jolla hosts an own instance (I took a look at that a long time ago: Not as shiny as Transifex, but definitely usable), is there a public instance you would expect its hoster to run it sustainably (i.e., potentially "for ever" & to perform proper maintenance, i.e., updating Pootle reasonably (no betas, no long outdated versions) etc.)?

nephros:
No, not pootle, Weblate.

I have one project there (ThemeColor), and they do indeed have a "Free Plan" for FLOSS projects, for exactly one project. (I therefore couldn't add this one there).

I guess they make money through AI training using the community translations. They also have a feature for automatic suggestions which use Deepl, Bing, Google and others so they presumably receive their share ;).

PR #48, opened 2021-02-18
Sfos4 fix

CODeRUS:


benaryorg:
I haven't gotten an answer in the [Jolla Forum thread](https://forum.sailfishos.org/t/patchmanager-patches-in-koli-4-0-1/4925/104) in three weeks so I'll comment here directly; the build artifacts did expire last month, and in lieu of other packaging (i.e. OpenRepos) the current way of using Patchmanager on 4.0 is compiling the code oneself, pulling sources directly out of the branch used for the PR no less. It would be greatly appreciated if mainline packaging had compatible versions, especially considering the patches are applied in a volatile nature with a way to circumvent the entire process effectively making the update non-lethal so to say.

CODeRUS:
uploaded here: https://github.com/sailfishos-patches/patchmanager/releases/tag/3.0.1

Phoenix616:
Patchmanager itself seems to work fine for me on 4.1.0.24, finding patches that are updated/still work with that version however is a different issue...

Olf0:
@CODeRUS, while discussing Pull Request #49, [you asked why I believe that this Pull Request here breaks compatibility to older SailfishOS releases](https://github.com/sailfishos-patches/patchmanager/pull/49/#issuecomment-878632524).
Sorry that it took so much time to answer, but I had looked over this PR #48 only briefly, was sure that there is at least one incompatible change, but wanted to check thoroughly that my assessment is exhaustive (i.e., that there are no incompatible changes I had not detected).

Now I am pretty sure that the only incompatible change is per commit https://github.com/sailfishos-patches/patchmanager/commit/8d974de4b63a6f75c5d9e984051606cb3beb57ca in [line 57 of `patchmanager-daemon.pro`](https://github.com/sailfishos-patches/patchmanager/blob/8d974de4b63a6f75c5d9e984051606cb3beb57ca/src/bin/patchmanager-daemon/patchmanager-daemon.pro#L57) (see [diff to previous commit here](https://github.com/sailfishos-patches/patchmanager/commit/8d974de4b63a6f75c5d9e984051606cb3beb57ca#diff-605f898d93c595811100256c6be266486ae27a623fe9f620f4e8ad7c19e251b3)): Changing `systemd.path` from `/lib/systemd/system/` to `/usr/lib/systemd/system/`.
But I am stating the obvious.  Plus, because SailfishOS does not provide an compatibility mechanism for this and a "Pro" file it is not a place where one could easily detect and depend on the installed SailfishOS release (in contrast to, e.g., RPM spec files), I see no easy way to avoid this incompatibility, except by maintaning two different Patchmanager branches for SailfishOS releases before and after this filesystem layout change.

PR #50, opened 2021-07-12
remove usage of `${PIPESTATUS[0]}` which doesnt work with busybox

b100dian:


Olf0 (approved):
Correctly altered syntax, so both shell scripts now work fine and do the same when executed either with bash or busybox's ash (which is symlinked to `/bin/bash` in recent SailfishOS releases per `busybox-symlinks-bash` RPM).
Thanks!

Olf0:
#### Side note
I still wonder ([as mentioned before]()), why @CODeRUS explicitly made the pipes return an exit code (by `; test ${PIPESTATUS[0]} -eq 0`) but does not check for it in most of the places, for example [line 95 of `pm_apply`](https://github.com/sailfishos-patches/patchmanager/blob/patchmanager3/src/tools/pm_apply#L95) and [line 82 of `pm_unapply`](https://github.com/sailfishos-patches/patchmanager/blob/patchmanager3/src/tools/pm_unapply#L82).
I also fail to understand on first sight why the few places in these two scripts, where the return code of pipelines is checked (in `pm_apply` only in the functions [`test_if_applied`](https://github.com/sailfishos-patches/patchmanager/blob/patchmanager3/src/tools/pm_apply#L55) and [`verify_text_patch`](https://github.com/sailfishos-patches/patchmanager/blob/patchmanager3/src/tools/pm_apply#L71), but never in [`pm_unapply`](https://github.com/sailfishos-patches/patchmanager/blob/patchmanager3/src/tools/pm_unapply)), it is done *within* the functions: These functions set the exit status of the piplines as their return code (because it is the return code of the last command executed in these functions), so it could be nicely evaluated in the shell script body ("main programme") at the bottom of the scripts.

I am writing this here, because I wanted some feedback, if I understood this correctly, before I start writing a two Pull Requests:
1. One not altering the current behaviour, which just moves the existing two checks in `pm_apply` to the main programme (after the function calls).
2. Another one adding additional checks, where I think it is appropriate (which I still have to evaluate properly).

What do you think?

b100dian:
> the pipes return an exit code (by `; test ${PIPESTATUS[0]} -eq 0`)

I think the pattern used throughout was just so you'd not get the exit code from `tee`.
Regardless if the exit code was actually used or not.

Olf0:
Nice, that you have been provided with commit rights in this repo!

You may consider to merge my long-standing, trivial documentation update MR #45 

My MR #46 needs a little review (see [this MR comment](https://github.com/sailfishos-patches/patchmanager/pull/46#issuecomment-709575128)), and @mentaljam's MR #44 looks fine for me and implements a long wanted "feature" (correct sorting), but my general coding skills are too low to properly review it.
It would be nice, if you find the time to review them, and merge them or ask for changes.

Olf0:
> Yes @Olf0 I will try to look at those 

Thank you!

> do you folks in this org have some other channel where you get together?

Well, you already know and use the communication channels GitHub provides (issue discussions and MR comments), plus FSO threads (for example the one for [Patchmanager and SFOS 4.0.1](https://forum.sailfishos.org/t/patchmanager-patches-in-koli-4-0-1/4925)).
Additionally, you may use TMO (talk.maemo.org), which also has a [Patchmanager thread](https://talk.maemo.org/showthread.php?t=92935)
Both FSO and TMO also allow for sending private messages (PMs).

But if you think of a common tool for chats: Not really, there are disjunct groups using different tools.  Many Russian users prefer Telegram (IIRC there is a Patchmanager group there), others prefer IRC (there is at least one SFOS specific group there), I prefer XMPP ("Jabber").

Issue #60, opened 2021-09-23
Is the syntax of lines 50 & 51 in the spec file correct?


Olf0:
Lines 50 & 51 in the [spec file](https://github.com/sailfishos-patches/patchmanager/blob/master/rpm/patchmanager.spec#L50-L51), which are currently
```
/usr/lib/qt5/bin/qmlplugindump -v -noinstantiate -nonrelocatable org.SfietKonstantin.patchmanager 2.0 %{buildroot}%{_libdir}/qt5/qml > %{buildroot}%{_libdir}/qt5/qml/org/SfietKonstantin/%{name}/plugin.qmltypes |:
sed -i 's#%{buildroot}##g' %{buildroot}%{_libdir}/qt5/qml/org/SfietKonstantin/%{name}/plugin.qmltypes
```
do not make sense to me:
1. In line 51 (the `qmlplugindump` call) 
   * The trailing `|:` must be wrong, because `qmlplugindump`'s output to std-out must be redirected to a file via `>`, so an additional redirection via a pipeline is nonsense, plus the bash built-in `:` does not read anything from std-in.
     Hence I assume it was supposed to be `||:`, which I would write as `|| true`.
   * I was not able to determine from [`qmlplugindump`'s documentation](https://doc-snapshots.qt.io/qtcreator-4.0/creator-qml-modules-with-plugins.html#generating-qmltypes-files), in which cases it may fail (making a `||:` or `|| true` necessary), but it states "*You can safely ignore the debug output.*", which means adding a `2> /dev/null` in order to discard it here (i.e., in a spec file).
   * Additionally I do not comprehend why `qmlplugindump` is called with then `-v` flag, but then its debug output is not used.
     For `qmlplugindump`'s source code and calling conventions, see https://code.woboq.org/qt5/qtdeclarative/tools/qmlplugindump/main.cpp.html#744
   * Furthermore that line is too long IMO, so this flaw at the line's end is not easily noticed.
2. Line 52 is supposed to process the freshly generated file, eliminating the prefix `%{buildroot}` for every path listed in it.
  ~~But I have never seen `sed`'s substitute command to work with the delimiter `#` (instead of `/`), and its man-page also does not list `#` as a special character.
  Am I completely missing some aspect (e.g., something crazy like `rpmbuild` transforms `#` which are not at the start of a line to `/`), or is this simply wrong?~~
[...]

PR #61, opened 2021-09-23
Increase version number & add dep to SFOS >= 3.4.0

Olf0:
Increase version number & add dep to SFOS >= 3.4.0
- Increase version number from 3.0.1 to 3.1.0
- Add runtime dependency to SailfishOS >= 3.4.0, because I always hated that Patchmanager was installable on old SailfishOS releases, which it does not support any more.

### DIFF
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -8,7 +8,7 @@
 Name:       patchmanager
 
 Summary:    Patchmanager allows for managing Sailfish OS patches
-Version:    3.0.1
+Version:    3.1.0
 Release:    1
 Group:      Qt/Qt
 License:    TODO
@@ -16,6 +16,7 @@ URL:        https://github.com/sailfishos-patches/patchmanager
 Source0:    %{name}-%{version}.tar.bz2
 Requires:   unzip
 Requires:   patch
+Requires:   sailfish-version >= 3.4.0
 BuildRequires:  pkgconfig(Qt5Core)
 BuildRequires:  pkgconfig(Qt5DBus)
 BuildRequires:  pkgconfig(Qt5Qml)
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -48,7 +48,8 @@ rm -rf %{buildroot}
 
 %qmake5_install
 
-/usr/lib/qt5/bin/qmlplugindump -v -noinstantiate -nonrelocatable org.SfietKonstantin.patchmanager 2.0 %{buildroot}%{_libdir}/qt5/qml > %{buildroot}%{_libdir}/qt5/qml/org/SfietKonstantin/%{name}/plugin.qmltypes |:
+/usr/lib/qt5/bin/qmlplugindump -noinstantiate -nonrelocatable org.SfietKonstantin.patchmanager 2.0 %{buildroot}%{_libdir}/qt5/qml \
+> %{buildroot}%{_libdir}/qt5/qml/org/SfietKonstantin/%{name}/plugin.qmltypes 2> /dev/null || true
 sed -i 's#%{buildroot}##g' %{buildroot}%{_libdir}/qt5/qml/org/SfietKonstantin/%{name}/plugin.qmltypes
 
 mkdir -p %{buildroot}%{_unitdir}/multi-user.target.wants/
@@ -103,8 +104,7 @@ echo '%{_libdir}/libpreload%{name}.so' >> /etc/ld.so.preload
 if ! grep -qsF 'include whitelist-common-%{name}.local' /etc/firejail/whitelist-common.local; then
    echo 'include whitelist-common-%{name}.local' >> /etc/firejail/whitelist-common.local
 fi
-dbus-send --system --type=method_call \
---dest=org.freedesktop.DBus / org.freedesktop.DBus.ReloadConfig
+dbus-send --system --type=method_call --dest=org.freedesktop.DBus / org.freedesktop.DBus.ReloadConfig
 systemctl daemon-reload
 systemctl-user daemon-reload
 systemctl restart dbus-org.SfietKonstantin.patchmanager.service
@@ -139,8 +139,7 @@ echo "Updating %{name}: postun section"
 ;;
 *) echo "Case $* is not handled in postun section of %{name}!"
 esac
-dbus-send --system --type=method_call \
---dest=org.freedesktop.DBus / org.freedesktop.DBus.ReloadConfig
+dbus-send --system --type=method_call --dest=org.freedesktop.DBus / org.freedesktop.DBus.ReloadConfig
 systemctl daemon-reload
 systemctl-user daemon-reload
 


Olf0:
Fix issue #60
- Rectify line 51: Note that all errors are suppressed in line 51, but line 52 fails, if line 51 did not successfully create that file.
- Beautify lines 107 and 142

Olf0:
Actually I would like to have these two commits *not* squashed.

Even generally, I do not see any advantage in squashing unrelated commits:
- Yes, that results in a shorter history, but the history also becomes unreadable.
- Additionally it prevents cherry picking from development branches (currently we have nine) or reverting single commits (because they might be squashed together with other, unrelated commits).

Hence my suggestion is to set the default back to a plain "Merge", and "Squash and merge" can still be selected, if the commits are related (classically, a base commit and enhancements / corrections of it).

Conflicting opinions (@b100dian)?

nephros:
Agreed, for the reasons stated. Squashing must be decided case by case.

But also PRs should ideally not contain separate unrelated changes in the first place. If they do, rewiewers may request a split into separate PRs.

I don't care which is the default, as long as the merger takes care to do the right thing. I agree with Linus though that Githubs making merge commits per default is annoying.

nephros (approved):

b100dian:
@Olf0 use whatever you think it's best for the PR - you have the three options enabled (maybe you can try the Rebase one, it shall fast-forward). The default is the last one used.

Olf0:
> [...] The default is the last one used.

Ah, this is why I did not find a default setting in the repository settings.
Thanks!

b100dian:
I've added a [3.1.0-pre tag](https://github.com/sailfishos-patches/patchmanager/releases/tag/3.1.0-pre) (I need a tag for publishing to chum) and used the Releases interface to pen some changelog, What did I miss? 
(Am I again too excited or aren't we at the first milestone:D?)

nephros:
Woohoo!

Let's cork the pops!  🍸🍹🍷

One rhing I'm not sure of: AFAIK things built on OBS/chum have the RPM vendor set to something that makes it *not* update things built elsewhere.

So if we publish the OBS-built package on Openrepos, pkcon/zypper will not update from older version.

If that is indeed so, which build artefacts do we choose to release  on Openrepos? OBS? GH Action? Something else?

Olf0:
> [...] maybe you can try the Rebase one, it shall fast-forward

BTW, rebasing can send one to Git-hell easily!
AFAIK, "Rebasing and merge" should only be done when absolutely necessary (i.e., the base branch was altered, so that one's changes do not apply cleanly any more) and handled with a lot of care, because ... it may send the repo to Git-hell.
If a commit-set does not apply cleanly (per usual merge or squash-merge), it is usually better do determine which commit in the base branch is causing this and to pull that into the commit-set, so it does apply cleanly, again.  At least that is my experience.

Olf0:
> I've added a [3.1.0-pre tag](https://github.com/sailfishos-patches/patchmanager/releases/tag/3.1.0-pre) (I need a tag for publishing to chum) and used the Releases interface to pen some changelog, What did I miss?

AFAICS, nothing. :smile:

> (Am I again too excited or aren't we at the first milestone:D?)

Well, coming close to it.

> One thing I'm not sure of: AFAIK things built on OBS/chum have the RPM vendor set to something that makes it not update things built elsewhere.
> So if we publish the OBS-built package on Openrepos, pkcon/zypper will not update from older version.

I do not believe that, but cannot disprove it either.
I played around with the vendor string a couple of times (for various reasons)in the past and all libzypp based tools (pkcon, zypper etc.) do not seem to care at all, what is is set to.  Why should they?
Hence "maybe there is an undocumented, special vendor string preventing updates of RPMs with any other vendor string" sounds almost like a conspiracy theory to me.
But as an engineer, my line of thinking is usually: Try to find out.  Someone has to install-test the 3.1.0-pre RPM, anyway.

b100dian:
> Let's cork the pops! 🍸🍹🍷

Now that's the attitude! :+1: :)

> Well, coming close to it.

Yes!

I also [remember](https://forum.sailfishos.org/t/community-obs-refurbished-and-re-floated/6556/23) problems installing from OBS (my initial build) that may apply to the chum build too.

I just tested the armv7hl version upgrade from OBS to https://github.com/sailfishos-patches/patchmanager/actions/runs/1270940701 and it works.

nephros:
I was mainly referring to this: https://forum.sailfishos.org/t/announcing-sailfishos-chum/7381/7

> Chum apps/packages should not overwrite your installed packages as they have chum set as vendor. So, at least zypper needs zypper in -f harbour-app-name to consider changing vendor of the package. Same goes other way - if you installed from chum, it should stay there. Don’t know about pkcon handling of it, though.

... and have observed the situation that:

1. enable repo A
2. install a package version 1.0 from repo A
3. enable repo B (chum)
4. have a version 1.1 available inchum
5. do pkcon refresh; pkcon update

in this case I have seen pkcon NOT picking up the updated package from chum. However, I have not thoroughly tested it, it may just be that it doesn't install newer versions from other repos, independent from the vendor tag.
But Rinigus' post seems to imply that this tag *does* have *some* effect. But is might just affect packages from jolla/jolla store, not at all sure.

b100dian:
@Olf0 you edited my comment instead of replying :joy:. I think you kept my words as quoted though.

I was going to update the chum version so that people that went on and installed pm3 from chum would have access to the update.

> Well, coming close to it.

@Olf0 do you still want to do the OBS publish? - and if so, what do you think it's missing?

@nephros nice digging onto the vendor issue! So if I understand correctly, in our case this will apply to chum vs openrepos - what I don't know if this is baked inside the package or just the repo's metadata (most probably the latter).

Olf0:
> @Olf0 you edited my comment instead of replying. :joy:

Argl, I wanted to "Quote reply" and hit "Edit" instead.
Oops, sorry!  I was not aware that I can do this.  To quote Andrey (I don't know who the original source is): "With great powers comes great responsibilities!" and it is unfortunate I one is not even aware of these powers.

>  I think you kept my words as quoted though.

Mmmh, probably not everything.  My bad!  I was cycling 85 km since then and do not remember at all what I might have discarded.

> @Olf0 do you still want to do the OBS publish? - and if so, what do you think it's missing?

No, I never cared specifically about OBS for generating the RPM.  The GH-artefact is sufficient for uploading at Openrepos, correct?
Or did you men "at Openrepos"?  Then my answer is, "no I have not gotten around to create that Openrepos page, and it will be at least late next week or nest weekend, until it is ready (first half of the upcoming week is too packed)".  And I should create the FSO page concurrently, so I can switch them on at the same time (at FSO: draft -> publish; at Openrepos: Library (not shown) -> App)

> @nephros nice digging onto the vendor issue

+1 (I already did a "thumbs up")

Olf0:
I also have two more things to clarify on my radar:
1. License: I do not want to publish anything without a license!  In most jurisdictions that is equivalent to put it into the public domain, which provides authors and users with basically no protections at all.  E.g., a company may take this, use and sell it, and if anything goes wrong, sue us, because we were distributing a defective product.  Any OSI-approved license is much better, even the BSD-style ones (including MIT and Apache license).
If I don't publish, I don't really mind, but advise everybody not to do this.
2. WRT the README.md here: Does PM3.x really still use AUSMT (as PM1 and 2 did)?  I assumed that the method first demonstrated by @jakibaki's pre-patch superseded the AUSMT-mechanisms, when PM3 (-beta) was created, correct?
But to resolve this is unrelated to "officially" publish PM 3.1.0.

b100dian:
Oops, I meant openrepos not OBS, yes.

nephros:
> I also have two more things to clarify on my radar:
> 
> 1. License: I do not want to publish anything without a license!  In most jurisdictions "no license" is equivalent to put something into the public domain, which provides authors and users with basically no protections at all.  E.g., a company may take this, use and sell it, and if anything goes wrong, sue us, because we were distributing a defective product.  Any OSI-approved license is much better, even the BSD-style ones (including MIT and Apache license; but I think we do have a preference for reciprocal licenses, as the *GPL*).

Most of the source files carry a 3-Clause BSD license header, none of them carry other license information.
Confirmation of the authors is good, but I think for now I think we have an answer to that question, no?

Olf0:
> Most of the source files carry a 3-Clause BSD license header, none of them carry other license information.

Peter, thank you for researching this more thoroughly than I did: I took a few samples (2 -3) before and somehow managed to only hit files without a license.
Triggered by your assessment, I re-checked, which fully confirms your observation.

> Confirmation of the authors is good, but I think for now I think we have an answer to that question, no?

Well, this statement carries three aspects IMO (even though you likely only intended a single one): :wink:
1. "I think for now I think we have an answer to that question."
    Yes, many 3-clause BSD licensed files, some without a license, overall license "ToDo". :stuck_out_tongue_winking_eye:
2. "for now"
    This is actually why I am pursuing this *now*!  The more contributors, the harder it becomes to switch to another license.
3. "Confirmation of the authors is good"
    Well, legally, it is necessary.

I really would like to switch to a *reciprocal* FLOSS-license, preferably (to keep it simple) to the `LGPL-2.1-only`.
If you and @nephros vote against that, I can live with 3-clause BSD for all files, even though I do not like the aspect of all *permissive* FLOSS-licenses, that people / companies may use this code in proprietary projects or products.

AFAICS, the unlicensed files (e.g., `pm_apply` and `pm_unapply`) are all written by @CODeRUS.
Hence I really want to see a clarification from him (and preferably also @SfietKonstantin) WRT
a. a license change of all files to `LGPL-2.1-only`: {`O.K.`|`No`|`Don't care`}
b. if the original intention was to put all files under `3-clause BSD`: {`O.K.`|`No`|`Don't care`}

If we get answers (at least from @CODeRUS), I will create a MR properly implementing either option a. or b.
But the current situation is not a proper basis to proceed with a public PM3.1 release for me. :slightly_frowning_face:

BTW, when looking up the SPDX identifier for `BSD-3-Clause`, I noticed that the [license text there](https://spdx.org/licenses/BSD-3-Clause.html) and also the (identical) [license text at the OSI](https://opensource.org/licenses/BSD-3-Clause) is slightly different at start of the third clause](https://github.com/sailfishos-patches/patchmanager/blob/master/src/qml/PatchObject.cpp#L16).
Never mind, its meaning is unchanged and it is clearly a '[BSD 3-Clause "New" or "Revised" License](https://opensource.org/licenses/BSD-3-Clause)', because all other of [the many BSD 3-Clause variants (or any other BSD license variant) ](https://spdx.org/licenses/) do differ significantly.

Olf0:
@CODeRUS, I do not comprehend your absolute silence on the topic of licensing!
After having addressed this per direct mail to your gmail account weeks ago, in multiple threads here, plus easing to answer as much as possible (e.g., a "*b.: O.K.*" would be fully sufficient), I can at least tell for sure that this behaviour is intentional and selective (towards me and / or the topic).
As much as I did want to get an answer and would like to understand your selective muteness, I cease to care.
But I must clearly state that this behaviour is frustrating, annoying and angering.

P.S.: And it feels so Jolla'ish.

PR #63, opened 2021-09-27
[pm_apply / pm_unapply] Add missing quoting

Olf0:
... plus a single superfluous quoting of a fixed string for consistency between pm_apply and pm_unapply

### DIFF
--- a/src/tools/pm_apply
+++ b/src/tools/pm_apply
@@ -125,7 +125,7 @@ log "PM APPLY $(date)"
 log "----------------------------------"
 log
 
-log $(basename "$PATCH_DIR")
+log "$(basename "$PATCH_DIR")"
 
 if [ -f "$PATCH_PATH" ]; then
   log "  contains text patch"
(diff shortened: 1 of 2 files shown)


Olf0:
Note: These two commits shall be squash-merged.

nephros (approved):

CODeRUS:
just curious: what the point of these changes? looks useless for me :))

Olf0:
> just curious: what the point of these changes?

Just "quoting best practices" applied on-the-go while I was looking for something else.

> looks useless for me :))

Well, in this very specific case they might be, functionally.  But only because of
```
log() {
  echo "$@" | tee -a "$PM_LOG_FILE"
}
```
so it stubbornly consumes all parameters per "$@".

But such constructs `var=$(basename "$path")` (without outer quoting) are the reason, why so many things fail, when paths or filenames contain spaces.

And if you think as an attacker, such missing quoting is often an invitation for code injection: Let the filename be "; rm -rf /" (idiotic example, late at night; but I think you get the point).
I learnt to always pay extra attention (to quoting etc.), when processing data, which is externally controlled (here: an arbitrary file-name)

TL;DR: If I see such pattern, where quoting should be applied, I just stubbornly do it, if I can.  Rather once too often, than once too less.

P.S.: And quoting the fixed string in `pm_unapply` was just a beautification and for constitency, because exactly the same variable assignment was already using quotes in `pm_apply` and *all* other variable assignments in both files do.

Issue #33, opened 2018-11-17
[PM3 suggestion] Retrieve the installed SFOS version by using 'version', not 'ssu re'


Olf0:
**Prelude:**
As some Patch authors fail to update the information, which SailfishOS releases a specific version of a web-catalog Patch is compatible to (even though that version is compatible to newer SFOS releases), users have practically no choice but to enable Patchmanager's *Developer mode*.

**Observed behavior:**
While the update notifications for web-catalog Patches (per system notification, in Patchmanager's top pulley, in PM's list view of the web-catalog and the Patch specific sub-page there) are working as expected, when a device is on the current SailfishOS release.
But when a device has an older SFOS release installed, some web-catalog Patches will have newer versions, which are only compatible to more recent SFOS release(s).  One surely does not want to update these Patches.
Still Patchmanager 3.0.55 notifies one via all aforementioned ways, that updates are available.

**Suggestion to mitigate this:**
When checking for updated Patches from the web-catalog, compare the currently installed SailfishOS release (e.g. per `version | rev | cut -f 2 -d ' ' | rev`, but not per `ssu re`) with the compatibility information of a Patch, for which an update candidate (i.e. newer version) exists, before generating notifications (i.e. do not notify, if the currently installed SFOS release is not supported by the update candidate).
IMO, this extra check always shall be active, regardless of Patchmanager's *Developer mode* being switched on or off.

Side note: This suggestion is only about when the update notifications are triggered.  The ability to install any version of a Patch, when Patchmanager's *Developer mode* is active is not at all addressed (and shall be retained).
<br />

Tested on a Jolla 1 phone under SailfishOS 2.2.1.18 (while SFOS 3.0.0.8 has been released a few weeks ago) with Patchmanager 3.0.55.

P.S.: As issue reports and feature suggestions implicitly always have a slightly negative touch (as proposing to better something), an explicitly positive statement: 
Kudos for implementing the multiple, aforementioned ways of notifying that there are update candidates for web-catalog Patches.  This is extremely helpful, already.

Olf0:
Oh, pardon me, Patchmanager 3.0.55 does all that almost fine!
[...]

Issue #64, opened 2021-09-28
[Sandboxing] Review the way things work in the light of full application sandboxing


nephros:
This is upcoming:

https://blog.jolla.com/whats-up-with-sandboxing/

We should collect how this will affect us in this thread.

Things to look at:

- [ ] calling DBus services we do not ship
- [ ] reading/writing files outside of `/usr/share/patchmanager` (e.g. `/etc`)

Actions to take:

- [ ] create an application profile for patchmanager

CODeRUS:
hope jolla-settings will have enough permissions :)
in worst scenario making separate desktop file for patchmanager :)

b100dian:
Hopefully sandboxing only applies for applications launched with the desktop file - and the Settings plugin should only talk through DBus with patchamanger-daemon, which we ship.., hopefully that's still possible.
Need to have 4.3 to test though.
Then, if Storeman can still install RPMs we should be good (famous last words)

CODeRUS:
sure, and settings for sure will have permissions to talk with dbus :))

Olf0:
See also [point 2 here](https://github.com/sailfishos-patches/patchmanager/pull/65#issuecomment-930535505) for my contribution to this discussion.

Plus mind that Andrey usually "knows & has future" WRT SFOS.

b100dian:
> With the upcoming Sailfish OS 4.3.0, you can also test if your application works with the default profile. You can enable the default profile by adding a file called “50-enable-sandboxing.conf” in /etc/sailjail/config, with the content:
```
[Default Profile]
Enabled=true
```

Just enabled this, all seems good. All apps started to require microphone/bluetooth/camera/internet/location/media index/NFC/play and record audio/removable media/user directories and web contet, but jolla-settings did not.

Olf0:
Oh, IMO this needs some pondering how to proceed procedurally:
As discussed at https://github.com/sailfishos-patches/patchmanager/discussions/113#discussioncomment-1582443 I believe / prefer to "single out" critical changes.  I think switching on the sandboxing this is one, but may be convinced of the opposite.
But if we agree that this may have unforseeable side-effects, I would like to do a 3.1.1 release without sandboxing activated, a 3.1.2 with it and ultimately a 3.2.0 release with the "32 /64 bit mangling" merged.

b100dian:
Sandboxing above was not enabled for patchmanager. We are using the jolla-settings desktop file I believe.
[...]

PR #65, opened 2021-09-28
Do not call ssu to determine OS Version

nephros:
See issue #33
	https://github.com/sailfishos-patches/patchmanager/issues/33

	Concept ported from @elros, ac0991b
	https://github.com/elros34/patchmanager/commit/ac0991b1b27df010066fa4e3f8f287c6b1004a3e

### DIFF
--- a/src/qml/WebPatchPage.qml
+++ b/src/qml/WebPatchPage.qml
@@ -326,7 +326,7 @@ Page {
                     width: parent.width
                     contentHeight: filesContent.height
                     property bool isInstalled: !!container.versions && container.versions[modelData.project] == modelData.version
-                    property bool isCompatible: modelData.compatible.indexOf(PatchManager.ssuVersion) >= 0
+                    property bool isCompatible: modelData.compatible.indexOf(PatchManager.osVersion) >= 0
 
                     onClicked: {
                         if (!PatchManager.developerMode && !isCompatible) {
--- a/src/qml/patchmanager.cpp
+++ b/src/qml/patchmanager.cpp
@@ -162,22 +162,7 @@ PatchManager::PatchManager(QObject *parent)
 
     });
 
-    QDBusPendingCallWatcher *watchGetSsuVersion = new QDBusPendingCallWatcher(m_interface->getSsuVersion(), this);
-    connect(watchGetSsuVersion, &QDBusPendingCallWatcher::finished, [this](QDBusPendingCallWatcher *watcher){
-        watcher->deleteLater();
-        QDBusPendingReply<QString> reply = *watcher;
-        if (reply.isError()) {
-            qWarning() << Q_FUNC_INFO << reply.error().type() << reply.error().name() << reply.error().message();
-            return;
-        }
-
-        qDebug() << Q_FUNC_INFO << reply.value();
-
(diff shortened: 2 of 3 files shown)


Olf0:
Even though you stated to "[not speak Qt and C++](https://github.com/sailfishos-patches/patchmanager/issues/33#issuecomment-928904955)", you seem to be able to write and express what you want properly.  I can barely comprehend most of that, thus a big thank you for adapting elros34's code for PM2.

> [Code seems fine, but](https://github.com/sailfishos-patches/patchmanager/issues/33#issuecomment-928886289):
> 1. this bit, while valid I assume, seems unrelated to the overall change:
   https://github.com/elros34/patchmanager/commit/ac0991b1b27df010066fa4e3f8f287c6b1004a3e#diff-da6a1e0621bec30e25650901e8a69669572d302bbccc7f82e5fc81f4b90e4e68L312
> 2. Thinking of the upcoming sandboxing, how do we feel about reading files from /etc instead of dbus-calling a system service? There are rumors that the latter might break when sandboxing is turned on fully, and the former will break certainly if the app profile doesn't allow it. (There's a separate [issue \#64 on that](https://github.com/sailfishos-patches/patchmanager/issues/64) though.)

* For point 1, I also believe it is unrelated.
* For point 2: Because there is not definite information, how these restrictions are applied, my stance is: "do not care about rumors"; lets test, when the code is out (or more specific information is provided, but that is purely theoretical with Jolla).

But for a proper review, I kindly ask @b100dian to do this, because I can only provide uneducated statements (e.g., "The code is looking beautiful, like a dense forest of words; and with syntax hilightning it all becomes so colourful!" :wink:).

b100dian (commented):
Changes look good, I just didn't test them.
In case there are problems with reading directly that file we can always add a dbus method on patchmanager-object to make the query for the UI plugin.

Olf0:
> Changes look good, I just didn't test them.

Mmmh, I always reviewed changes by re**viewing** them. :wink:
More seriously, I never tested any changes up to now, before setting "Approved", just reviewed them thoroughly.  I may test, if I am very curious / I deem a change very critical / I have time to spoil.

So I do / did not think that any kind of testing is required for approving a MR.
I also believe we should not set the quality-assurance and procedural hurdles too high, as that may stall processing MRs etc.

Do our opinions on this differ (@b100dian sounds a bit like they do)?
Are there some reviewing guidelines, policies or best practices I am not aware of (and which include some testing)?

In this specific case (and many others, when there are non-trivial changes to QML or C++ code), I cannot seriously review due to my lack of knowledge and experience.  And I sure will not just hit "Approved" to let things proceed.

So @b100dian, it is either you or potentially nobody who approves this MR.

b100dian (approved):
This could have been breaking things, that's why the test - I don't test all PRs, of course. The visual ones if I'm more curious:)
Just did a quick run on this and it's fine.
I wasn't able to trigger the original bug just by running `ssu re 4.2.0.20` (e.g. change to an non-existant version)

nephros:
Damnit, too soon!

src/bin/patchmanager-daemon/patchmanagerobject.h/cpp must be updated to use the new osVersion actually, otherwise compatibility check on the web catalog does not work.

PR #70, opened 2021-09-29
Show warning label for conflicting patches

nephros:
Add a (Unicode) warning character to patch names if they might conflict with others.
Also fixes spelling of the corresponding warning.

### DIFF
--- a/src/qml/PatchManagerPage.qml
+++ b/src/qml/PatchManagerPage.qml
@@ -445,7 +445,7 @@ Page {
                     anchors.right: patchIcon.status == Image.Ready ? patchIcon.left : parent.right
                     anchors.margins: Theme.paddingMedium
                     anchors.verticalCenter: parent.verticalCenter
-                    text: name
+                    text: ( patchObject.details.conflicts.length > 0 ) ? name + " ⚠": name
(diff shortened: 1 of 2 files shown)


b100dian (approved):
👍

Olf0:
Oh, please don't ... too late.
See, the conflict detection results in many, many false positives, because it works quite coarsely by detecting, if two patches patch the same file.
I once asked for enhancing this to check more precisely (i.e., to check if two patches try to patch the same lines in the same file), because I and others were annoyed by the many false positive statements "Confilcts with:".  If that is implemented, a warning sign is fine.
With the currently implementation IMO no such a warning should be shown and rather the test message also be softened to "May conflict with:"!

TL;DR: It is of no use and just irritating the users, if an exponentially growing number of patches are displayed with a warning sign, when one installs more and more patches.

Olf0:
Thanks!
Much more reasonable now in your nephros/patchmanager branch.

How is this planned to update the code here?

nephros:
> Thanks! Much more reasonable now in your nephros/patchmanager branch.
> 
> How is this planned to update the code here?

Phew, glad you like it! :D 

I don't think it's necessary to revert PR #70, I'll just put a PR from that branchch on top of it as usual.

PR #73, opened 2021-10-02
Rework PR-70: Show warning label for conflicting patches

nephros:
This reverses/reworks #70 and improves the patch info page a bit:

 - remove the "warning sign" character from patch list
 - show number of conflicting patches in Patch List context menu
 - list conflicting patches in Patch Info page
 - show applied state of patch in above list
 - patch log could previously be copied to clipboard by tapping on it. But there was no hint that this was possible, or that it happened. Added a notification about it, and a hint that you can do that.


nephros:
While updating the LegacyPatchPage with the same changes, I noticed the differences aren't that big.
So I removed both of them, and created UnifiedPatchPage instead.

src/qml/UnifiedPatchPage.qml
```diff
@@ -68,84 +83,89 @@ Page {
...
-            Label {
-                color: Theme.highlightColor
-                anchors.left: parent.left
-                anchors.right: parent.right
-                anchors.margins: Theme.horizontalPageMargin
-                wrapMode: Text.WrapAtWordBoundaryOrAnywhere
-                text: modelData.compatible.join(', ')
+                horizontalAlignment: Text.AlignJustify
+                font.pixelSize: Theme.fontSizeSmall
+                text: qsTranslate("", "This patch uses the legacy format for its patch.json file. If you're the maintainer consider updating to the new format. If you are using Web Catalog you should not include a patch.json file at all in your upload.")
```
b100dian, line 156: > If you're the maintainer consider updating to the new format. If you are using Web Catalog you should not include a patch.json file at all in your upload

Is this possible? This is the set of files extensions allowed: https://github.com/CODeRUS/django-test/blob/master/example/demo/filehandler.py#L29

src/qml/patchmanagermodel.cpp
```diff
@@ -239,6 +239,17 @@ QString PatchManagerModel::patchName(const QString &patch) const
     return m_patchMap[patch]->details()->value(QStringLiteral("display_name")).toString();
 }
 
+bool PatchManagerModel::isApplied(const QString &name) const
+{
+    // FIXME: there certainly is a more efficient way, e.g. std::find_if?
```
b100dian, line 244: If this was sorted, std::lower_bound :)
Would've been cool if there was already a map populated onPatchedChanged from JS directly - not sure if this is triggered inside the PatchObject::setData call.

(also that loop in PatchObject::setData is very weird. I almost filed a bug, it is like a tech debt, not actually a bug)

Anyway, I'm fine with this FIXME for the moment -> meaning that we should still find a better solution.

src/qml/qml.pro
```diff
@@ -26,11 +26,10 @@ DISTFILES += qmldir \
     AboutPage.qml \
     DevelopersPage.qml \
     RestartServicesDialog.qml \
-    LegacyPatchPage.qml \
+    UnifiedPatchPage.qml \
```
b100dian, line 29: :bow:

src/qml/PatchManagerPage.qml
```diff
@@ -480,10 +480,11 @@ Page {
                         }
                     }
                     MenuLabel {
-                        visible: patchObject.details.conflicts.length > 0
-                        text: qsTranslate("", "May have conflicts")
+                        visible: !patchObject.details.patched && patchObject.details.conflicts.length > 0
+                        text: qsTr("May conflict with %n other patch(es), see %2", "", patchObject.details.conflicts.length).arg(patchinfoitem.text)
```
b100dian, line 484: Interesting use of translation parameter.. Another translation! Let's see how this parameter goes with translations, I have not seen any "Open the Patch info page" or something like that as `%2` as a quick check.

b100dian (approved):

src/qml/UnifiedPatchPage.qml
Olf0: As you have shown this is technically not possible.

Hence it is just the semantics of the human language (i.e., here: English)!
This is why I proposed [the use of RFC2119 language](https://github.com/sailfishos-patches/patchmanager/issues/78#issuecomment-934730025): Most "should" in Andrey's texts mean "must", though some of them really mean "should" and a few of them "may"!
The same with "shall", which sometimes is supposed to mean "must", but more often "should" (while it is definitely closer to "must" than "should" in English).

nephros: The patch that 'triggered' this addition and the reworking of the Legacy page is https://coderus.openrepos.net/pm2/project/email-list-view-line-count which *does* include a legacy patch.json despite being a rather new patch.  
How it can be there?

> > If you're the maintainer consider updating to the new format. If you are using Web Catalog you should not include a patch.json file at all in your upload
> 
> Is this possible? This is the set of files extensions allowed: https://github.com/CODeRUS/django-test/blob/master/example/demo/filehandler.py#L29

It absolutely is, I just tested.

WC does actually allow a json file in the uploaded file, and does not replace/mangle it if it exists. How/why that check is bypassed I don't know.

This means you can actually ship a json file with information different than what was entered in WC. You then can e.g. have your patch listed as compatible in the Web Catalog list (because there data from the WC api call is used), but after installing you have the patch show up as red/incompatible in the patch list, because then information from the json file is used.

nephros: Thanks for your review and changes. Much appreciated.  I shall try follow the style you propose in future changes.

> As you have shown, this is technically not possible. Hence it is simply subtly wrong use of semantics of a human language (i.e., here: English), hence the "should not" must be a "must not"!
> 
>[...]
> 
> Suggestion to resolve this here: Delete the whole third sentence, which cannot be the case, any way. 

But I have just shown, (if only to myself but anyone can easily verify), that it is actually possible to include any json file with the package you upload at WC. So the language describes the real-world situation even if it may be against the 
 "specification".

I would like to keep the sentence in because I believe that
 - we should encourage moving to the new format, because it has most importantly compatibility information which the legacy has not
 - the "nag" on the patch page is probably the place most visible to patch devs, who might not follow any changes on our docs site.

src/qml/UnifiedPatchPage.qml
```diff
@@ -32,14 +32,29 @@
...
     property var delegate
+    property bool legacyPatch: !modelData.isNewPatch
     signal doPatch
 
+    Notification {
+        id: popup
+        appName: modelData.display_name
+        summary: qsTranslate("", "Log copied to Clipboard!")
+        previewSummary: summary
+        icon: "image://theme/icon-s-clipboard"
```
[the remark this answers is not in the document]
nephros: `appIcon` actually overrules `icon` if present, at least for a popup. I'll rip both of these lines out.

src/qml/PatchManagerPage.qml
nephros: Heheh, yeah let's see how this one goes.

And sorry for inconsistently using `qsTr()` instead of `qsTranslate()` here, I wasn't sure the latter actually can do this.

src/qml/UnifiedPatchPage.qml
b100dian: Just wanted to add that the test I linked to, while testing `.endswith(extensions)` passes because `".json".endswith(".js")`:)

b100dian (approved):
Updates look good

PR #75, opened 2021-10-03
Fix PR 64: Move from SsuRelease to OsRelease

nephros:
Should fix some fallout from PR #65, which removed Ssu calls from patchmanager, but left pachmanagerobject still using ssu values. 

This led to the compatibility checks on web catalog patch pages failing (all versions displayed as incompatible).


b100dian (approved):
Sorry I tested this without the version check in place (old developer mode) and didn't notice. Thanks for fixing it!

Olf0 (approved):
Looking good, AFAICT.

PR #79, opened 2021-10-05
Fix zip package handling, enhance decompression capabilities

nephros:
I noticed during dogfooding that downloading packages from Web Catalog which have .zip packages would not work.

The download would complete successfully, but then the patchmanager daemon would start unzipping, and the unzip child process would hang indefinitely (apparently waiting for input on fd 0).  
This then causes the patchmanager daemon to hang while it waits.

Suspecting this might have to do with zip packages being placed in `/tmp`, I moved that location within `/tmp/patchmanager3/work`, and changed the call to unzip to force-overwrite files in the target dir. This makes unzipping work again, although I'm not exactly sure why.

This PR also includes two other enhancements:

 - make all paths to external binaries static constants on the top of the source file
 - support more than just .tar.gz tarball packages (tar.bz2, tar.xz...) - although Web Catalog doesn't support those at the moment.

Please review (for clarity, lookat the commits individually) and let me know:

 - do you see the unzipping problem with current PM as well? Does this fix it for you if yes? The example package that showed the problem for me is https://coderus.openrepos.net/pm2/project/email-list-view-line-count
 - do you find the additional tarball support useful? I can rip it out again as it is not useful at the moment.

### DIFF
--- a/src/bin/patchmanager-daemon/patchmanagerobject.cpp
+++ b/src/bin/patchmanager-daemon/patchmanagerobject.cpp
@@ -107,6 +107,13 @@ static const QString AUSMT_INSTALLED_LIST_FILE = QStringLiteral("/var/lib/patchm
 static const QString PM_APPLY = QStringLiteral("/usr/libexec/pm_apply");
 static const QString PM_UNAPPLY = QStringLiteral("/usr/libexec/pm_unapply");
 
+// external binaries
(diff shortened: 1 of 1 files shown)


b100dian (approved):
Looks good - and we can add tar.bz2/xz to web catalog later

Issue #78, opened 2021-10-05
Update Documentation on creating patches, and patch.json formats


nephros:
1. Documentation on how to create a new patch is sparse and could be improved. [1], [3] Especially the correlations of packaging patch.json and using a Web Catalog upload:
  - patch.json ins *only* required for RPMs (or more generally patches not distributed via Web Catalog)
  - patch.json should use the newer format (but old is supported)
  - Web catalog creates a patch.json from the info entered in the web interface and packages it automatically

2. patch.json contents are not documented at all (AFAIK)

3. Some of the documentation points to [2] as an example for a patch.
While this still works as an example, it uses the legacy patch.json format, which has much fewer information bits than the newer one.

References:
[1] Web catalog "Documentation (kinda)" page: https://coderus.openrepos.net/pm2/usage/
[2] Example project pointed to by README currently: https://github.com/CODeRUS/sailfishos-disable-glass-avatar
[3] Project README: https://github.com/sailfishos-patches/patchmanager/blob/master/README.md
[4] probably various guides floating around at the usual places (TJC, TMO, etc.)

Olf0:
#### Side note 1
There is also [5] https://openrepos.net/content/coderus/patchmanager-30
But that does not seem to add anything significant over [2]. 

#### Side note 2
Well, point 2 is documented in [4]:
```json
{
    "name": "My super patch",
    "description": "Some description.",
    "category": "other",
    "infos": {
        "maintainer": "Foo Bar"
    }
}
``` 
#### Main message
I think the primary task is to consolidate [2] and [4] into [4] (mind that I inserted a pointer to a subsection) and adapt to proper specifcation language (following [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119) closely, but not necessarily its spelling in all capital letters): No "shall", no "can"; only "must", "should" and "may", plus their negations.

Reasons: [4] already is the richest source of information and it is the easiest to discuss and work on, because it it hosted at Github.

Olf0:
https://github.com/sailfishos-patches/patchmanager/pull/73#discussion_r722775922 is a good example for making [3] the principal documentation and the only place where it is maintained.

PR #81, opened 2021-10-05
Update documentation

nephros:
Implement some if the things mentioned in Issue #87

https://github.com/sailfishos-patches/patchmanager/issues/78


Olf0 (approved):
Looking O.K., but need another review (because I changed a lot), while reading issue #78 first.

Also: Which other "archive files" (name formats!?!) in [line 18](https://github.com/sailfishos-patches/patchmanager/pull/81/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R18)

b100dian (approved):
big :+1:

Olf0:
Did I correctly add a trailing comma in https://github.com/sailfishos-patches/patchmanager/pull/81/files#diff-4e19fb89ae9a5a8f061f2ccba71814601d00d625a7f836333b1c983a11d953fcR7 ?

nephros:
> @nephros, did I correctly add a trailing comma in line 7: https://github.com/sailfishos-patches/patchmanager/pull/81/files#diff-4e19fb89ae9a5a8f061f2ccba71814601d00d625a7f836333b1c983a11d953fcR7 ?

Absolutely! Good catch, I spotted two or three others now. Fixed.

Olf0:
I now also deleted this (merged -> stale) branch, because I already merged stuff on top of it in master.

Olf0:
> This is a good discussion to have.
> My feeling right now is that I am "an elephant in a china store" because I have got admin rights to this org and the transifex one.
> Therefore, before I get comfortable with all the things I could break, I would like to remain a single elephant (not three:) and discuss changes like this.

Yes, let us have this discussion.  Originally I intended to open a new discussion thread for this, but simply continuing here is fine for me.

I am also glad, that you seem to be receptive to open (and from my side always clear, sometimes bordering "harsh") statements.

I cannot concur, for a couple of reasons, which fall into two categories:
practical and social considerations

Under the social aspects, I can tell that it felt strange that my repeated offer to Andrey to take care and responsibility for this Github repo and the Transifex repo (explicitly including to responsibly distribute administrative rights) was met with absolute silence, but ultimately you received administrator rights for both after triggering him via Telegram.  I do not know and stopped to care what his considerations were; except if he instructed you to be restrictive with sharing administrative rights (i.e., please state if he wanted that).

Nevertheless, this creates a situation, in which we are not a team of peers (i.e., "on a par" / on equal level; which does not mean "be equal", because everyone has different strengths).  E.g., I have to ask you (as I already did a couple of times), if you transfer the necessary rights to me to perform an action or perform the action by yourself.
Specifically assuming the "paper pusher" role means that you are the only one to judge and to ultimately decide.  The same goes for your statement that you prefer to be the only "Elephant" who has the ability to break things.  This is basically self-assigning the project leader role.
While that clearly puts @nephros and me in a subordinate role, my primary concern is why you apparently expect him and me not to alter settings and to perform administrative work as carefully as you do?

Furthermore this is a consideration I really cannot follow, as it was rather me questioning the need and suggesting a conservative / conservational approach to all significant or disruptive changes (e.g. deleting branches) plus double checking that such actions do not discard anything which might be of value (even if that is just for keeping history documented).
And I cannot see that any action by @nephros or me indicated, that one of us may not responsibly and carefully handle administrative powers.
If you perceive one of these two points differently, I would really like to know that and what triggered that perception.

Under practical aspects this approach does not make any sense to me:
* It creates extra work and delays, because @nephros and I simply cannot perform some tasks, as e.g. inviting or accepting new translators at Transifex.  Instead we have to ask you to do that.  This feels awkward in a team of volunteers, creates the additional work to explicitly ask each time and increases your workload with administrative stuff.
* You are a single point of failure: I.e., in case something unfortunate happens (illness, accident etc. or even just "severely out of time") nobody else has the administrative rights to fully substitute you.
  Exactly this is the reason why I asked for a secure communication channel to share the password for the "Patchmanager" account at Openrepos.

<br/>
<br/>
For the original trigger of this, the issue template:
It was freshly created (for FSO), simply creates a pre-filled page when opening a new issue (no one is forced to use it) and such guidance for bug reporters has shown to greatly enhance the quality of bug reports.
Just take a look at old, closed issues: Most lack the basic set of information as Patchmanager version and SailfishOS version.
Hence I cannot (and do not) really believe that this is questionable under practical aspects.

b100dian:
I've created patchmanager-admins group. Do you see the whole settings now @Olf0 and @nephros ?
I've also added Peter on transifex as Project Maintainer. I will upgrade you both to Organization level when I found where that is.

BTW, we should probably release those ~4 translation updates.

Olf0:
> I've created patchmanager-admins group. Do you see the whole settings now @Olf0 and @nephros ? 

Yes, I do: Thanks.

> I've also added Peter on transifex as Project Maintainer. I will upgrade you both to Organization level when I found where that is (DONE).

> BTW, we should probably release those ~4 translation updates.

I still have to check the remaining changes from PR #46, now that I may be able to do that WRT access rights.

Please slow down, 4 translation updates are not sufficient new material to trigger a new release IMO.  Ususally there shoiuld be at least a bug fix or some functional enhancement or change.
Maybe having carried out all these maintenance tasks would suffice:
* Getting the whole translations stuff in order and having checked that.
* Update donations page.
* Eliminate the [choppy scrolling of the About page](https://github.com/sailfishos-patches/patchmanager/pull/68#issuecomment-941837751) (with or without the easter-egg).

Olf0:
> Deleting branches on merge was something I thought you would appreciate (it sounded like you need that) so, agreeing, I said it's 2/3 rule.

Well, I hardly can express clearer than that, [that I neither "need" or want it](https://github.com/sailfishos-patches/patchmanager/pull/81#issuecomment-939298802):
> I do not like that, [...] Automatisms can turn against you!
> Please set it back to its original setting.

@nephros, we do need your opinion on the "autodelete branch after merging" function!

--------------------------------------------------------------------
@nephros, we also need your opinion on the topic issue temlpate.
* I [suggested one here](https://github.com/sailfishos-patches/patchmanager/pull/81#issuecomment-939298802), which is [the one used at FSO](https://forum.sailfishos.org/t/bugs-in-patchmanger-3-1-0/8552/2).
* @b100dian [argued against it](https://github.com/sailfishos-patches/patchmanager/pull/81#issuecomment-939369389):
  > In this specific case, I don't currently think an issue template will help ourselves (as I've already added note-to-self bugs and you folks added all kinds of scratch-an-itch issues) and since we're not invaded with "unable to reproduce" bugs, i don't think it will help users.
* I [argued for it](https://github.com/sailfishos-patches/patchmanager/pull/81#issuecomment-939380944):
  > It was freshly created (for FSO), simply creates a pre-filled page when opening a new issue (no one is forced to use it, as it can be freely altered, including deleting it completely per \<Ctrl-A\> & \<Del\>) and such guidance for external bug reporters has shown to greatly enhance the quality of bug reports at many projects.
Just take a look at old, closed issues: Many lack the most basic information as Patchmanager version and SailfishOS version.
Hence I cannot (and do not) really believe that this is questionable under practical aspects.

----------------------------------------------------------------------
I think we should also establish another rule, covering the case when we do not have an odd number of opinions:
Stick to the default setting (because that is usually reasonable).

So if @nephros does not voice an opinion, this would result in no autodelete and no issue template.

Olf0:
> First, being a data hoarder myself, I share @Olf0's instinct, but the branches get deleted only after a merge and thus no commits are lost. Given this, I believe @Olf0 should name some of the reasons for keeping merged branches around. The remark
> 
> > I have no idea how long [branches] can be restored
> 
> seems to betray some kind of misunderstanding of the branch concept in Git: @Olf0's concerns might be unfounded. A git branch is just a label attached to a commit and can be "restored" at any time:
> 
> ```
> git branch $name $hash
> ```
> 
> is all it takes and there's no clock ticking.

And where do the content of $name and $hash come from?

Look, this discussion is primarily about Github's webfrontend, so my statement "*I have no idea how long [branches] can be restored*" was meant to address only that.  IIRC this option there ("Restore branch") vanishes after some time.
It is about:
* Responsibility: IMO someone (usually the creator of that branch) should actively delete that branch as a consciously taken last step after merging.
* Choice: For more complex Pull requests, I like to keep the corresponding branch for a few days or weeks, just in case something should arise for which comparing, copying&pasting etc. at the web-frontend comes handy.
* Convenience: For all tasks which can be handled at the web-fontend, I do like to handle them there and want to keep this possibility open for others.  

BTW, it is not about "being a data hoarder", because these temporary branches shall be deleted some time after merging; but IMO not automatically and immediately. 

> Second, I believe that a canonical repo which receives contributions from multiple developers and strives for openness like sailfishos-patches/patchmanager, should not receive direct pushes from contributors. See, if you guys push your branches into this repository, you'll either have to let every J. Radom Hacker do the same, or you'll have a very two-tier development group. If I had some changes for patchmanager today, I'd have to submit them as a pull request from roman-neuhauser/patchmanager:some-branch -- and that's perfectly fine! It doesn't require any privileges on my part, it's open, and it scales to any number of contributors. Not to mention the potential for naming conflicts when multiple people push feature branches into the canonical repo: it's asking for one person to obliterate another's work. All it takes is two branches named something like `doc-update` and a blip of less-than-100% concentration.

1. I do not think there is anything wrong with a "two-tier development group" (why the "very"?!?), i.e. having a "core team" of developers proven to act responsibly and to back this project in the long run.  IMO it is rather a natural setup, and it was a long way to create a team with shared responsibilites and mutual trust.
I do see any reason why "open to contributions" shall mean "everybody contributes the same way", and actually no advantages from this approach, but some disadvantages.
I cannot see why this approach "does not scale" and way it would have to (it is only the core team members), and branch naming conflicts never occurred and are easily avoided by using GitHub's web-frontend.
Plus it does not help with any of my aforementioned points of the "autodelete function", even though you think otherwise.

2. But it would make close collaboration in the core team harder, especially when reviewing Pull Requests: This way one can simply add commits.
Just to anticipate you stating that there are settings, which enable this for branches in private repos (e.g. "allow for commits from maintainers" at the web-frontend): They have to be deliberately set and I do not like (just a gut feeling) to alter others people's repos.
Again, I cannot see any advantage, but do see disadvantages when switching to such a setup.

Olf0:
BTW @b100dian, I was able to address [your concerns WRT an issue template](https://github.com/sailfishos-patches/patchmanager/pull/81#issuecomment-939369389):
  > In this specific case, I don't currently think an issue template will help ourselves (as I've already added note-to-self bugs and you folks added all kinds of scratch-an-itch issues) and since we're not invaded with "unable to reproduce" bugs, i don't think it will help users.

Because I did not set a default template, but just created a template in `/.github/ISSUE_TEMPLATE/`, a new issue is not pre-filled with the template: Instead one chooses between an empty issue (as usual) or the template(s) as a first step when creating a new issue.

The issue template mechanism is a GH functionality, which turned out to be surprisingly nice: Simple and mighty at the same time, multiple different, templates can be offered and pre-filling new issues with one is optional.

PR #109, opened 2021-10-15
Convert patches between 32 and 64 bit library path

### LINKED ISSUE #71: Patches with targets in /usr/lib[,64] do not apply on 32/64bit systems
Systems with 64bit userland use `/usr/lib64` instead of `/usr/lib`as a library path.
Patches which target files in either of those locations will therefore not apply on the other platforms even though the patched files themselves might be the same.

This affects e.g. all "silica" patches, as the Sailfish QML files live under `$LIBDIR/qt5/qml/Sailfish/*`, and basically anything QML that isn't applications in `/usr/share`.

This means patch developers must currently ship and deploy two versions of each patch, and create two projects on web catalog, or use RPM packaging which installs the appropriate patch version.

Question:

Shall patchmanager implement something that helps with this problem?

Solutions that come to mind:

 - use something like `patch -p3 --dry-run -d /usr/lib{,64}/` when a `/usr/lib` string is detected in the patch file
   - probably the simplest and fastest way, but is it reliable?
   - breaks when files outside of `/usr/lib` are patched from the same diff file
 - accept a second diff file in the patch hierarchy (e.g. `unified_diff_64.patch`) so devs can ship either or both
   - this would require modifcation of both patchmanager and the Web Catalog validator for submitting tarfiles
 - modifying Web Catalog to mangle any uploaded diffs to create the missing variant, and supply the correct tarball when PM downloads, so PM always receives a valid patch depending on bittiness
   - wonky, as the catalog can't verify that a file will exist on the system
 - modify `/usr/lib*` paths in existing unified_diff files when installing a patch from Web Catalog 
 - modify `/usr/lib*` paths in existing unified_diff files when applying a patch at runtime 
 - faking the libpath (force /usr/lib also on 64bit) somehow when applying a patch
   - the last three would allow old 32bit patches to be used without the authors needing to update them
 - Other ideas?

nephros:
Fixes #71
https://github.com/sailfishos-patches/patchmanager/issues/71


nephros:
As discussed, we don't want to duplicate the mangle list in both cpp source and shell script.
So introduce a config file to hold that information, and read it from both.

nephros:
@b100dian please review, seems to work but may be bad c++.

b100dian:
Nothing wrong per se in the changes (just nits) but, however I tried playing around with 1. having only one MANGLED_CANDIDATES entry and 2. see if pm_unapply would work and even 3. If a setting could show the paths to the user (not clear how to pass the setting to pm_apply now, maybe `$2`) in my [fork](https://github.com/sailfishos-patches/patchmanager/compare/patch-bitness...b100dian:patch-bitness?expand=1) - but currently I get errors applying the Return old pulley menu 4.x that used to work. So I need to figure out if any of (my|your) changes did this last since I tested this. That's for tomorrow..

b100dian:
Ok @nephros managed to pinpoint it to toStringList not splitting the QVariant value - I have made a PR against this PR for the changes are somewhat numerous. Let me know if you can review that and merge --squash it. There is one caveat, see https://github.com/sailfishos-patches/patchmanager/pull/140

b100dian (approved):
All good from my point of view, maybe it needs more testing. Oh, and the wording in the settings page maybe needs review.
 
I've added a last commit refreshing patches on bitness mangling change and fixed the 'last' caveat.

b100dian:
Hmm maybe I spoke too soon, I still get errors like "PatchManagerObject::tryToUnlinkFakeParent" `error while trying to cd from "/usr/lib" to: "qt5"` on dynamic switch

b100dian:
Ok, just tested this build with 4.3.0 armv7hl and the setting does not harm existing 32-bit patches, and it tried to fix this 64-bit patch https://coderus.openrepos.net/pm2/project/patch-five-cameras-tucana which didn't apply as it shouldn't, but for the right reasons (the  `/etc/dconf/db/vendor.d/jolla-camera-hw.txt` file was not having the expected contents)

Other than this, with the last change (re-order to mangle before test if applied) it seemed to work correctly even at pm3 upgrade scripts on my 4.2.0 64-bit device.

b100dian:
Zig-zagging through my notifications, I need to test the last two checkboxes. Could find some time tomorrow.

b100dian:
Tested by "does it work reliably with patches that contain both lib64 and lib paths (e.g. patches a Silica file and a systemd service on 64bit)?" appending this to an existing patch
```
--- /usr/lib/patchmanager-test/test.txt
+++ /usr/lib/patchmanager-test/test.txt
@@ -1 +1,2 @@
-# Test line
+# test
+# replaced line
--- /usr/lib/test.txt
+++ /usr/lib/test.txt
@@ -1 +1,2 @@
-# Nothing
+# test
+# replaced line

```
To an existing patch, looks alright (it patched one file in lib64 and another one in lib)

PR #118, opened 2021-11-01
Adapt to current set of (top-level) laguages

Olf0:
See https://www.transifex.com/coderus/patchmanager3/languages/


Olf0:
I wonder, if we should rather use the "Send to GitHub" function at https://www.transifex.com/coderus/patchmanager3/settings/integrations/ , which does exactly the same as using this shell script plus manually posing a PR does: Checking languages for "100% translated" and posing a PR for those.
Side note: ... with its current settings which are a bit hidden: At the aforementioned page press "⋮" → "Edit settings" → Scroll all the way down in the settings window, which pops up.

P.S.: I will try that today, if it works as intended.

Olf0:
> P.S.: I will try that today, if it works as intended.

It does work fine, one can set a percentage threshold for the completeness of each translation to be submitted (nice!).
The only drawback is, that a separate PR is created for each translation.  OTOH, this way they can be reviewed and merged individually, which is a plus.

PR #136, opened 2021-11-01
Show list of things to be restarted on the "Restart Services" dialog

nephros:
"Inspired by" Issue-23


nephros:
Again @b100dian please check my CPP, I'm sure it's horrible.

b100dian:
> Again @b100dian please check my CPP, I'm sure it's horrible.

I actually find Qt to be very powerful for not tripping on horrible C++ things and is really nicely used by you:) I'll look in depth ~tomorrow

I need to ramp up on dbus types, but recently I found that `as` works directly as `QStringList` and have not found a need for the annotation (nor do I know what `a{s}` is by just reading https://dbus.freedesktop.org/doc/dbus-specification.html )

I'll try to look tomorrow

nephros:
> I need to ramp up on dbus types, but recently I found that `as` works directly as `QStringList` and have not found a need for the annotation (nor do I know what `a{s}` is by just reading https://dbus.freedesktop.org/doc/dbus-specification.html )

I came up with it more or less trial-and-error, and IIRC I got compilation errors when trying 'as'. It's possible something else was the root cause for those though.

src/qml/patchmanager.cpp
```diff
@@ -98,6 +98,22 @@ PatchManager::PatchManager(QObject *parent)
 
     });
 
+    QDBusPendingCallWatcher *watchGetToggleServicesList = new QDBusPendingCallWatcher(m_interface->getToggleServicesList(), this);
+    connect(watchGetToggleServicesList, &QDBusPendingCallWatcher::finished, [this](QDBusPendingCallWatcher *watcher){
```
b100dian, line 102: I get no clue from the documentation that this fires any more than once. 
I think it is also true that patch updates are notified only when entering the PM settings page and the Restart Services is finicky in the sense that it doesn't show if you first open the menu. I am not 100% sure as I haven't debugged this.

For example, I had to close and re-open the Settings app to get more than one item in the RestartServicesDialog Repeater.

So I guess the call for watcher can be moved to the getter?

b100dian: Actually this call is useful in the qml page, so it could be made synchronous with no devastating consequences. 
The repeated query needs to be done for the boolean that makes the menu entry appear.

Btw, thanks, I did not know what Restart services did until today!

src/qml/RestartServicesDialog.qml
```diff
@@ -39,21 +43,44 @@ Dialog {
...
+        }
+        SectionHeader { text: qsTranslate("", "List of services:" ); color: Theme.secondaryHighlightColor }
+        Column {
+            id: col
+            anchors.horizontalCenter: parent.horizontalCenter
+            width: parent.width -  Theme.horizontalPageMargin * 2
+            Repeater {
+                model: PatchManager.appsToRestart
+                delegate: Component { TextSwitch {
+                    text: modelData //TODO: this displays the raw strings from the daemon. Should be formatted/enabled for translation
```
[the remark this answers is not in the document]
Olf0: IMO they must be.

src/qml/patchmanager.cpp
nephros: > I get no clue from the documentation that this fires any more than once.
Hmm, that's suboptimal ;)

> I think it is also true that patch updates are notified only when entering the PM settings page and the Restart Services is finicky in the sense that it doesn't show if you first open the menu. I am not 100% sure as I haven't debugged this.
> 
> For example, I had to close and re-open the Settings app to get more than one item in the RestartServicesDialog Repeater.

The easiest way to get it to appear is to unapply and re-apply a patch (or preferrably, more than one from different categories).


> So I guess the call for watcher can be moved to the getter?

I'll play around with that and see how it goes. THanks for the thorough testing.

src/qml/RestartServicesDialog.qml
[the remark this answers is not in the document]
nephros: See ce498ad

src/qml/RestartServicesDialog.qml
```diff
@@ -39,21 +43,44 @@ Dialog {
...
+        }
+        SectionHeader { text: qsTranslate("", "List of services:" ); color: Theme.secondaryHighlightColor }
+        Column {
+            id: col
+            anchors.horizontalCenter: parent.horizontalCenter
+            width: parent.width -  Theme.horizontalPageMargin * 2
+            Repeater {
+                model: PatchManager.appsToRestart
+                delegate: Component { TextSwitch {
+                    text: qsTranslate("Sections", modelData)
```
nephros, line 70: This works for all languages but English, because English doesn't have a dedicated translation and so uses the (lowercase) original strings.

src/qml/patchmanager.cpp
nephros: okay, moved to getter - seems to work.

@b100dian as I didn't completely get your second comment - is that what you had in mind?

src/qml/RestartServicesDialog.qml
nephros: Nah, totally minor issue that just came up along the way.

src/qml/patchmanager.cpp
b100dian: Yes, that is exactly what I was suggesting - a plain synchronous getter. I am not sure about "seems to work" - did you also see problems with the one-off triggered callback from the original version (and existing pattern)?

(While I usually do prefer asynchronous calls, this may not be one of the cases to sweat on).

src/qml/RestartServicesDialog.qml
b100dian: Thanks for fixing this!

b100dian (approved):

src/qml/patchmanager.cpp
nephros: > Yes, that is exactly what I was suggesting - a plain synchronous getter. I am not sure about "seems to work" - did you also see problems with the one-off triggered callback from the original version (and existing pattern)?

Not really. One thing I'm not sure about is Signals. Do we need one, when to fire it, when/how it is used form QML when constructing/reading the model etc.

src/qml/RestartServicesDialog.qml
Olf0: Just to document that Transifex uses "translated" strings from the source file as source strings, when the are available (which makes sense IMO).
Thus this change changed the source strings of all 14 PM-categories for all 14 translations.  And because Transifex does not allow programmatic changes at Transifex (e.g., per RegEx), I manually adapted these 14 terms for 13 translations (all, but zh).  Maybe scripted downloading, programmatic changing the strings locally and scripted uploading would have been easier, but that would have required learning how to use Transifex's web API, which I wanted to avid this time.

Olf0: @nephros, please use the exclamation mark only if something went wrong.
Unfortunately the classic conventions for writing dialogues are less obeyed than in the early 2000s, but I still would like to stick to them.
* **!** means: Attention, something went wrong.
* **...** (concluding, i.e., at the end of a phrase or sentence) means: A new window is going to be opened.   This is most suitable for choices (which would open a new window).
* There are more, I do not remember spontaneously.

The dialogues in MS-Windows and most of KDE and Gnome are adhering to these conventions, so people are used to them (even most not consciously).

src/qml/patchmanager.cpp
b100dian: I think that since the model is not changed without QML calling the getter we won't need the `toggleServicesListChanged` and QML will only call the getter once when the dialog is shown up (or: every time).

So basically if the `appsNeedRestart` is working correctly for the menu to appear, the getter will too.

(I still don't have an explanation why `appsNeedRestart` works more than once though. Maybe it so happens that it is only set from false to true only once and from there the getToggleServices is called when you navigate to the page)

b100dian:
Just re-tested this and it works nicely

Olf0 (approved):
Looking fine.

Olf0:
> So, would you like this in now, or do we wait until after 3.2?
> 
> @Olf0 as you have taken the mantle of release manager for the next couple of versions, 

Huh, have I?
I merely intended to coordinate the releases a bit by asking who wants what in when.
In order to proceed in a structured manner and to avoid multiple, technically or functionally overlapping, major changes in a single release.

> please feel free to hit that green button whenever you think its appropriate.

Well, this has become surprisingly big by accumulating a lot of fix-ups, but is clearly disjunct (technically and by functionality) from the principal, major feature for the next release: "32 / 64 bit" per #71 and https://github.com/sailfishos-patches/patchmanager/pull/109#issuecomment-944535519 (which still has two checkboxes unchecked).

Hence I will happily merge this now.

PR #140, opened 2021-11-01
Patch bitness tweaks

b100dian:
This PR against `patch-bitness` branch

- reduces the MANGLE_CANDIDATES to only one list (32 bit) for cpp code too
- adds a dbus call for showing getMangleCandidates() in QML Settings
- adds patchmanager2 `bitnessMangle` setting which e.g controls `pm_apply` and cpp refresh list mangling
- favours the editied patch for unapply if present

Switching the bitnessMangle after applying a (failed) patch may not work without unapplying all patches.
Maybe we should mark this as experimental, but the reason is somewhere in the same cache preparation.


nephros (commented):
Very nice!

nephros (approved):

PR #144, opened 2021-11-02
Improve verbosity of log output

nephros:


Olf0:
So the aim is to adjust the verbosity:
- Be less verbose for the console output: console.log->info, log->debug, log->warn (seems that the "log" log-level was stubbornly used before)
- Mostly also decrease verbosity for Qt messages, but increase it in a few places from qDebug -> qInfo and qWarning -> qCritical

Looks like you checked each place well, and currently PM is overly verbose at the command line and in the systemd journal.  Hence I assume this is a nice improvement: Thanks!

Olf0 (approved):
Looking good.

nephros:
> P.S.: Although I have to admit to have no idea what `QT_LOGGING_RULES="*.debug=false"` in `src/bin/patchmanager-daemon/environment/10-dbus.conf ` does.

The QT_LOGGING_RULES env var controls the actual output (for any Qt application).

So you can select which categories you want to see by setting it to e.g. `QT_LOGGING_RULES="*.debug=false;*.warning=false;*.info=true` to shut up some categories.

Running e.g. `QT_LOGGING_RULES="*.debug=false" jolla-settings` will shut up patchmanager even without the changes in this PR.

PR #150, opened 2021-11-04
Increase version to 3.1.1

Olf0:
For details, see https://github.com/sailfishos-patches/patchmanager/discussions/113

### DIFF
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -8,7 +8,7 @@
 Name:       patchmanager
 
 Summary:    Patchmanager allows for managing Sailfish OS patches
-Version:    3.1.0
+Version:    3.1.1
 Release:    1
 Group:      Qt/Qt
 License:    BSD-3-Clause


nephros (approved):
OK, but I guess this is the last step before release. Might warrant a tag too.

Olf0:
> OK, but I guess this is the last step before release.

Yes, that was my intention.
With a CI running IMO this is the only reasonable way to handle version increases: As the very last commit before letting the CI build the release.  If something goes wrong and a new commit followed by a new build is needed, but the version number should stay the same, the release number in the spec file must be increased (*never* publish two RPMs with the same version-release string(s)).

Side note: Without using a CI I tend to increase the version number right after a release (to the next planned one).

> Might warrant a tag too.

Yes, I do tag the last commit for a release when creating it [at the web-frontend](https://github.com/sailfishos-patches/patchmanager/releases/tag/3.1.1).  That is exactly why I wanted this (the version increase) to be the last commit for this release.

I hope I interpreted correctly what you intended to express with the comment "Might warrant a tag too.".

PR #153, opened 2021-11-04
Translate Category names in ENglish

nephros:
... which just adds Capitalization

As we don't have a dedicated -en translation file any more because of
PR #108 , and we can't change the strings directly in the source, as

  - that would likely make the web catalog not work any more
  - would require changing of the .json specification
  - ^^^ would require patch authors to change all .json files
  - would require changing of the web catalog web backend

... this adds translations for strings used in the cpp source


Olf0 (approved):
O.K.

PR #185, opened 2021-11-05
Generate -en.qm file from translation source

nephros:
Follow-up to PR #153 

Up until now there was no .qm file installed for English (only existing -XX.ts files were considered by the build process).
In order to have translations show up we generate and install an -en.qm file, using the translation source as source.

### DIFF
--- a/patchmanager.pro
+++ b/patchmanager.pro
@@ -37,6 +37,11 @@ qm.commands += lupdate -noobsolete $${TRANSLATION_SOURCES} -ts $${TS_FILE} && \
 # create the qm files
 qm.commands += ; [ $$HAVE_TRANSLATIONS -eq 1 ] && lrelease -nounfinished $${TRANSLATIONS_OUT} || :
 
+# special case: as TS_FILE serves as both source file as well as
+# the English translation source, create the en qm file from it:
+qm.files += $$replace(TS_FILE, \.ts, -en.qm)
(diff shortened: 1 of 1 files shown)


Olf0:
> Follow-up to PR #153
> 
> Up until now there was no .qm file installed for English (only existing -XX.ts files were considered by the build process).
> In order to have translations show up we generate and install an -en.qm file, using the translation source as source.

Sounds reasonable, though I wonder what "*In order to have translations show up [...]*" means, because the translations are already working fine AFAICT.
But maybe I am just missing *where* the translations "show up" with this change.

nephros:
> > Up until now there was no .qm file installed for English (only existing -XX.ts files were considered by the build process).
> > In order to have translations show up we generate and install an -en.qm file, using the translation source as source.
> 
> Sounds reasonable, though I wonder what "_In order to have translations show up [...]_" means, because the translations are already working fine AFAICT. But maybe I am just missing _where_ the translations "show up" with this change.

Consider the latest change of "translating" the categories from Cpp files ("homescreen", "other" etc) to their capitalized equivalents.

Those "translations" would not show up for me when testing the list of services list PR #136 in English locale (still the lowercase versions show up).

The reason being, the runtime environment will look for a .qm file containing these translations - but while before this change here the build process *did* create a .qm file from the `settings-patchmanager.ts` file, it did not package/install it.

Olf0 (approved):
Looking good, though I am not fully able to check the correctness of the syntax (i.e., please double-check).

nephros:
> Looking good, though I am not fully able to check the correctness of the syntax (i.e., please double-check).

It does what it's supposed to do.  
There likely is away to do this more elegantly, but someone who writes .pro files more often than me is welcome to refactor the whole thing :)

PR #199, opened 2021-11-10
[ui] Add option to suppress notifications

nephros:
When applying a lot of patches the notification popups get in the way, and clutter the Notification view.

This allows the user to turn off success notifications for applying patches.

### DIFF
--- a/src/bin/patchmanager-daemon/patchmanagerobject.cpp
+++ b/src/bin/patchmanager-daemon/patchmanagerobject.cpp
@@ -219,15 +219,16 @@ void PatchManagerObject::notify(const QString &patch, NotifyAction action)
     qDebug() << Q_FUNC_INFO << patch << action;
 
     Notification notification;
-    notification.setAppName(qApp->translate("", "Patchmanager"));
-    notification.setHintValue("app_icon", "icon-m-patchmanager2");
-    notification.setTimestamp(QDateTime::currentDateTime());
 
     QString summary;
     QString body;
     QVariantList remoteActions;
 
     switch (action) {
+    case NotifyActionNone:
+        qDebug() << Q_FUNC_INFO << "Notification suppressed.";
+        return;
+        break; //well...
     case NotifyActionSuccessApply:
         summary = qApp->translate("", "Patch installed");
         body = qApp->translate("", "Patch %1 installed").arg(patch);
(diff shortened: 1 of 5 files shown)


src/bin/patchmanager-daemon/patchmanagerobject.cpp
```diff
@@ -1817,8 +1822,13 @@ void PatchManagerObject::doPatch(const QVariantMap &params, const QDBusMessage &
         }
     }
 
+    // is this used anywhere??
     if (!params.value(QStringLiteral("dont_notify"), false).toBool()) {
-        notify(displayName.toString(), apply ? ok ? NotifyActionSuccessApply : NotifyActionFailedApply : ok ? NotifyActionSuccessUnapply : NotifyActionFailedUnapply);
+        if (getSettings(QStringLiteral("notifyOnSuccess"), true).toBool()) {
```
b100dian, line 1827: suggestion: `if getSettings(QStringLiteral("notifyOnSuccess"), true).toBool() || !ok` would require one less enum and no return + break:p

b100dian: Wait, the default would be true?

Olf0: WRT the default: Please do not change the extant behaviour!
I.e., all these notifications shall be shown, as long as not deliberately switched off.

nephros: That was the thought behind that default, yes. Keep behaviour, add option to improve (IMO) behaviour.

b100dian: Sorry, true is correct in this case, of course. I got sidetracked.

nephros: > suggestion: `if getSettings(QStringLiteral("notifyOnSuccess"), true).toBool() || !ok` would require one less enum and no return + break:p

Yes - however that logic is (for me) already hard to parse.  
I mean

    if (!params.value(QStringLiteral("dont_notify"), false).toBool())

"if not value dont, (or false) then..." followed by a triple `a ? b : c`...

That return is ugly, yes but I think the enum entry maked it clearer, and it might be reused, e.g. when this is extended to not show other notifications (e.g. one might want to turn off update notifications).

nephros: OK "refactored" a bit according to your suggestion, it's clearer now I think.

b100dian: Ah, yes, no more double?-ternary?-senary? operator, I like it:)

b100dian (approved):

Olf0:
@nephros, AFAICS this is ready to be merged.

Olf0:
> @nephros, AFAICS this is ready to be merged.

Or is there any reason not to merge this yet?

nephros:
> > @nephros, AFAICS this is ready to be merged.
> 
> Or is there any reason not to merge this yet?

Just inertia, thanks for the reminder. Hitting the green button now! :)

PR #205, opened 2021-11-14
Update ts source file. Use "Convert Patches between 32 bit and 64 bit"

b100dian:


Olf0:
Also fixes a couple of incorrect line numbers in `translations/settings-patchmanager.ts`!

Olf0 (approved):
Looks O.K.

Issue #10, opened 2017-08-13
[Bug] Handling updated compatibility information of Patches


Olf0:
**Prelude:**
When SFOS 2.1.0.11 was released, most Patches in Patchmanager's Web catalog were still marked as for "... 2.1.0.10 EA" at most, so I had to toggle "Enable developer mode" in order to install them. This worked nicely and these Patches are displayed in red in Patchmanager's installed Patches list (which is good).
Meanwhile all these Patches were refreshed and are now marked as compatible to 2.1.0.11 in the Web catalog, but their version numbers were not increased (most of them are at 0.0.2). 
**Bug description:**
Consequently Patchmanager 2 obviously does not handle them as updated, hence still displayes them in red in Patchmanager's installed Patches list and still shows "Compatible: 2.0.1.10" in the detailed Patch information (by a tap on a Patch name) there. 
**Expected behaviour:**
Updates of the "Compatible with" information in the Web catalog should trigger an update of this information for already installed Patches from the Web catalog, thus eventually not displaying them in red anymore. 

Tested with Patchmanager 2.3.1-10.39.1.jolla on a Jolla 1 phone under SFOS 2.1.0.11

Originally posted [on TMO as issue 1.](http://talk.maemo.org/showpost.php?p=1527185&postcount=2208) due to issue #9 (filing this here [as promised](http://talk.maemo.org/showpost.php?p=1529055&postcount=2245)).

Olf0:
Retested with Patchmanager 3.0.55 under SFOS 2.2.0, 2.1.1 and 3.0.0 on Jolla 1 phones and Xperia Xs: Behavior unchanged.

Issue #17, opened 2017-08-13
[Usability bug: Inconsistent wording] "install*" instead of "apply*" and "remove" instead of "unapply" used in notifications 


Olf0:
**Description:**
The wording in Patchmanager2 consistently keeps "installing / uninstalling" and "applying / unapplying" a Patch clearly separate, which is excellent. But in the notifications "install*" instead of "apply*" and "remove" instead of "unapply" is used, which is confusing. 

**Possible solution:**
Please alter the three notifications messages, e.g. as follows.
Note, I also removed the word "Patch" and the action taken repeated in each second line, because the second line is usually way too long for fitting into the transient notification pop-up in the upper left corner of the screen, and the duration the transient notification is displayed is too short for the second line to automatically scroll horizontally (it usually just starts scrolling shortly before the transient notification vanishes).
a. Patch installed : Patch XXX installed
--> Patch applied : "XXX"
b. Failed to install patch : Patch XXX installation failed
--> Failed to apply Patch : "XXX" conflicts with ["YYY" | and "ZZZ" | original file <filename>]
c. When unapplying a Patch:
Patch removed : Patch XXX removed
--> Patch unapplied : "XXX"
d. But when uninstalling a Patch:
Patch removed : Removed patch XXX
--> Patch uninstalled : "XXX"
e. And it seems that:
Failed to remove patch : Patch removal failed
--> Failed to uninstall Patch : "XXX"
Or can Patches also fail to be unapplied (then one has to differentiate, as in c. & d.)?
f. Finally, the remorse timer label currently states
Uninstalling patch XXX in N seconds
which then ought be (just for consistency)
Uninstalling Patch "XXX" in N seconds

Tested with Patchmanager 2.3.1-10.39.1.jolla on a Jolla 1 phone under SFOS 2.1.0.11

Originally reported [on TMO as issue 8.](http://talk.maemo.org/showpost.php?p=1527989&postcount=2224) (due to issue #9; filing this here [as promised](http://talk.maemo.org/showpost.php?p=1529055&postcount=2245)).

Issue #11, opened 2017-08-13
[Usability issue] Missing ability to uninstall Patches in the web catalog (i.e. where they are installed)


Olf0:
As a Patch can be installed in a Patch details page within the Web catalog, I expected a possibility to uninstall them there as well. It took me quite a while to understand that both, Patches installed from Warehouse and Patches installed from the Web catalog, can be uninstalled by a long press on a installed Patch name in Patchmanager's installed Patches list.
Please provide a possibility to uninstall Patches in their Patch details page within the Web catalog as well (e.g. by a top pulley entry or by also a long press on an installed Patch version), as one expects an ability to uninstall, where one installs (this is already valid for RPM-Patches).
Still, I believe uninstalling by a long press on a installed Patch name in Patchmanager's installed Patches list should stay as well; once found, it is quite useful there and works for *all* Patches (regardless, if from Warehouse or the Web catalog). 

Tested with Patchmanager 2.3.1-10.39.1.jolla on a Jolla 1 phone under SFOS 2.1.0.11

Originally posted [on TMO as issue 1.](http://talk.maemo.org/showpost.php?p=1527185&postcount=2208) (due to issue #9; filing this here [as promised](http://talk.maemo.org/showpost.php?p=1529055&postcount=2245)).

PR #209, opened 2021-11-21
In-memory fix for compatibility update

b100dian:
This should fix https://github.com/sailfishos-patches/patchmanager/issues/10

However this only updates the in-memory model at check-for-updates. So it will 'work' only as long as you have once checked for updates successfully during a daemon instance run.

The patch.json file is not re-written with the new contents. 
What would the side effects of that be?

### DIFF
--- a/src/bin/patchmanager-daemon/patchmanagerobject.cpp
+++ b/src/bin/patchmanager-daemon/patchmanagerobject.cpp
@@ -2400,6 +2400,7 @@ void PatchManagerObject::requestCheckForUpdates()
 
                 QString latestVersion = patchVersion;
 
+                QStringList compatibleNow{}; // Used to detect changes in 'compatible' field.
                 const QVariantList files = project.value("files").toList();
                 for (const QVariant &fileVar : files) {
                     const QVariantMap file = fileVar.toMap();
@@ -2407,12 +2408,21 @@ void PatchManagerObject::requestCheckForUpdates()
                     if (!compatible.contains(m_osRelease)) {
                         continue;
                     }
+                    compatibleNow = compatible;
                     const QString version = file.value("version").toString();
                     latestVersion = PatchManagerObject::maxVersion(latestVersion, version);
                 }
 
                 if (latestVersion == patchVersion) {
                     qDebug() << Q_FUNC_INFO << projectName << "versions match";
+                    if (compatibleNow.length()) {
+                        if (m_metadata[projectName][COMPATIBLE_KEY] != compatibleNow) {
+                            qDebug() << Q_FUNC_INFO << projectName << "update the compatibility to " << compatibleNow;
+                            m_metadata[projectName][COMPATIBLE_KEY] = compatibleNow;
+                        }
+                        // This only marks the flag in-memory. The patch.json file is not re-downloaded.
+                        m_metadata[projectName][ISCOMPATIBLE_KEY] = compatibleNow.contains(m_osRelease);
+                    }
                     return;
                 }
                 qDebug() << Q_FUNC_INFO << "available:" << projectName << "version:" << latestVersion;


Olf0:
First of all: Thank you for taking care of this issue #10.
(Side note: When I manage to finalise addressing issue #17, we have addressed all my old issue reports but one: issue #11)

~~While I do see that this approach is nicely non-intrusive from a coding perspective, I believe that the resulting behaviour is too confusing for the average user.~~

Sorry, I have not taken the time to understand the original code, but it seems that the compatibility information is separate from the JSON file of a Patch (from the Web Catalog), so it can be and is separately downloaded and evaluated?
But it is also part of the JSON file proper, isn't it?
IMHO this design is confusing, the JSON files are small enough (a few hundred bytes to 100 kbytes) to simply handle them as an atomic unit, and re-download a Patch from the Web Catalog as a whole, when something changed (e.g., the compatibility information).

But it looks like you ultimately implemented this nicely WRT usability while retaining the original, slightly complicated design.

b100dian:
I'm not very happy with how things are:)
The initial non-intrusive patch was made under the observation that the check-for-updates only went for this response
`curl https://coderus.openrepos.net/pm2/api/project?name=patch-five-cameras-tucana | jq .`
(if you have jq installed to format it)

As you can see, `patch.json` is not entirely present there.

The workflow was: if `files` section indicated a newer version, notifiy that updates are available. Then the user would go and install the updated as would normally the first time.

The 2nd take was like: ok, I don't have the patch.json immediately available because we're just checking for updates. But, if the _current_ version has a newer/different compatibility field, let's store _just_ that silently.

It is still up for testing.. you need a patch to edit the versions back and forth and somehow trigger the check-for-updates (I think closing the Settings ui and re-opening it then going to Patchmanager)

b100dian:
Regarding downloading a fresh patch.json: maybe the whole patch could be re-downloaded if the (same) version changes. But then, maybe the author broke something, I wouldn't do that automatically.
Also, other fields in patch.json (or the `files` section in the above request) could be problematic to update automatically: what if the name changes or stuff like this? Or the section? What are the consequences?

This little hack should probably be more guarded around "only if the compatibility list added some new versions" too.

I'm open to ideas how to make this better (considering the code around is already a bit convoluted. Maybe I should bite the bullet and simplify some, but not today:)

Olf0:
> Regarding downloading a fresh patch.json: maybe the whole patch could be re-downloaded if the (same) version changes. But then, maybe the author broke something, I wouldn't do that automatically. Also, other fields in patch.json (or the `files` section in the above request) could be problematic to update automatically: what if the name changes or stuff like this? Or the section? What are the consequences?

Absolutely true (i.e., things are often not as simple as they seem)!
Thank you for clearly pointing this out.

Olf0:
> This little hack should probably be more guarded around "only if the compatibility list added some new versions" too.

Yes, if nothing was changed, except for some additional SailfishOS releases on the compatibility list, this approach is to be safe AFAICS.

nephros:
I'm still not clear on what the ideal workflow or algorithm should look like. AFAICS we have these cases:

 1. File version field bump. That's easy, higher version wins, -> notify user
 2. Except it's not so easy, higher *compatible* version (according to web catalog JSON response) wins. Or do we simply always notify, even for incompatible versions?
 3. Patch description `last_updated` field changed. not sure, probably ignore. Or trigger compatability check.
 4. File "uploaded" field: do we have something locally we can compare this to? Yes, the "last_updated" field of patch.json. But, when installing or updating a patch tar, is this field set to web catalog "last_updated" (patch) or "uploaded" (file) field?
 5. "Compatible" changed. When to check? Surely when file "uploaded" changed. Maybe when patch description "last_updated". Other times? Always for installed patches just to make sure?

Olf0:
@nephros, issue #10 is only about your case 5: Solely something in the `"compatible"` field for the installed version of a Patch was altered (usually new SailfishOS release(s) have been added), everything else is unchanged.
AFAIU (i.e., I have not checked that) only updating the `"compatible"` field for a specific version ogf a Web Catalog Patch does not update the `"uploaded"` field for that version or the `"last_updated"` field of the Patch.

> When to check?

Honestly, I have not explicitly thought about this before, but simply assumed, "when the Web Catalog is checked for updates".

b100dian:
Sorry for  the delay answering here. This PR only detects "same-version, but become-compatible" changes (or that's the goal). It runs at web catalog update check indeed. And it then only changes the compatibility of the stored patch json.
Now, I don't remember if last_updated or uploaded changed when doing tests for this, so I will need more time to check this.

PR #215, opened 2021-12-17
[shell programming] string comparisons vs. numerical ones / echo vs. printf

Olf0:
While quoting the getconf output may be overly cautious (and the output of other, external programs is still unquoted: grep, sed etc.),
using string comparisons (per test command / built-in) is much faster than numerical comparisons.

### DIFF
--- a/src/tools/pm_apply
+++ b/src/tools/pm_apply
@@ -15,7 +15,7 @@ PM_LOG_FILE="$PM_VAR_DIR/patchmanager.log"
 PM_PATCH_BACKUP_ROOT_DIR="$PM_VAR_DIR/patches"
 PM_PATCH_BACKUP_DIR="$PM_PATCH_BACKUP_ROOT_DIR/$1"
 
-SYS_BITNESS=$(/usr/bin/getconf LONG_BIT)
+SYS_BITNESS="$(/usr/bin/getconf LONG_BIT)"
 
 # Constants
 PATCH_NAME="unified_diff.patch"
@@ -76,7 +76,7 @@ test_if_applied() {
 
     $PATCH_EXEC -R -p 1 -d "$ROOT_DIR" --dry-run < "$PATCH_PATH" 2>&1 | tee -a "$PM_LOG_FILE"
 
-    if [ $? -eq 0 ]; then
+    if [ $? = 0 ]; then
       success
     fi
   fi
@@ -87,14 +87,14 @@ mangle_libpath() {
     if [ -f "$PATCH_PATH" ]; then
       log
       log "----------------------------------"
-      [ $SYS_BITNESS -eq 32 ] && log "Checking paths for 32->64bit conversion"
-      [ $SYS_BITNESS -eq 64 ] && log "Checking paths for 64->32bit conversion"
+      [ "$SYS_BITNESS" = 32 ] && log "Checking paths for 32->64bit conversion"
+      [ "$SYS_BITNESS" = 64 ] && log "Checking paths for 64->32bit conversion"
       log "----------------------------------"
       log
 
       found=0
       candidates="$MANGLE_CANDIDATES"
-      if [ $SYS_BITNESS -eq 32 ]; then
+      if [ "$SYS_BITNESS" = 32 ]; then
         # first, convert the candidate list
         # variable expansion ${foo/lib/lib64} would work on bash, but not POSIX/ash/busybox
         candidates=$(printf '%s' "$MANGLE_CANDIDATES" | sed 's@/usr/lib/@/usr/lib64/@g' )
@@ -103,11 +103,11 @@ mangle_libpath() {
       for p in $candidates; do
         totl_lines=$(grep -c "^+++" "$PATCH_PATH")
         cand_lines=$(grep -c "^+++ $p" "$PATCH_PATH")
-        if [ $cand_lines -eq 0 ]; then
+        if [ $cand_lines = 0 ]; then
           continue # nothing found, try next
         else
           found=$(( $found + $cand_lines ))
-          if [ $totl_lines -ne $cand_lines ]; then
+          if [ $totl_lines != $cand_lines ]; then
             # if there are several references in the patch file our mangling might do too much and cause the patch to fail.
             log "WARNING: mixed patch, conversion might not work"
           fi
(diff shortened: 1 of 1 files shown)


Olf0:
ToDo's:
- Check if the busybox's `ash` in SailfishOS is compiled with [`printf` as a built-in](https://github.com/brgl/busybox/blob/master/shell/ash.c#L107) (its default).  If not check if `echo` is and convert to it.
- ~~Check if path-hierarchy created by [`patch_edited_path="$PM_PATCH_BACKUP_DIR"/"$PATCH_EDITED_NAME"`](https://github.com/sailfishos-patches/patchmanager/pull/215/files#diff-1f9594d23b4b5870a4a217ad14d0793fa20ad488ecb80f6638be1cab1bf003f0R116) makes sense.~~  It does.
- Consider pulling [`mkdir -p`](https://github.com/sailfishos-patches/patchmanager/pull/215/files#diff-1f9594d23b4b5870a4a217ad14d0793fa20ad488ecb80f6638be1cab1bf003f0R119) out of the `for` loop.

nephros:
> ToDo's:
> 
> * Check if the busybox's `ash` in SailfishOS is compiled with [`printf` as a built-in](https://github.com/brgl/busybox/blob/master/shell/ash.c#L107) (its default).
>   If not check if `echo` is and eventually convert to it.

1. Yes it is
```
nemo@PGXperia10:~ $ busybox ash
nemo@PGXperia10:~ $ type printf
printf is a shell builtin
nemo@PGXperia10:~ $ echo $0
ash
nemo@PGXperia10:~ $ echo $BASH_VERSION

nemo@PGXperia10:~
````
2. `printf` is much, much more portable then `echo`. If anything, echo should be rmeoved from any and all scripts out there.

nephros:
> using string comparisons (per test command / built-in) is much faster than numerical comparisons.

Just out of curiosity, do you have a source for this? Because *usually* it's the other way around...

nephros:
>  using string comparisons (per test command / built-in) 

Thinking about this some more, I am against it.
We don't really have a perfprmance issue here, and string comparison is *unsafe* in the sense that

    a=2
    [ $a = "2 " ]     # is false, while  
    [ $a  -eq "2 " ] # is true.

so IMO arithmetic comparison is more robust (if a little less precise, but we're talkin *shell* here...).

Olf0:
Hi @nephros,

BTW, this was marked a a draft, i.e., "not ready for review".
Still some of your injunctions were really helpful:

> > Check if the busybox's `ash` in SailfishOS is compiled with [`printf` as a built-in](https://github.com/brgl/busybox/blob/master/shell/ash.c#L107) (its default).  [...]
> 
>     1. Yes it is. `echo` is as well.  [...]

Thanks. (Was a task I wanted to address now.)

>     2. `printf` is much, much more portable then `echo`. If anything, echo should be removed from any and all scripts out there.

Uh, this is a bold statement.
Once upon long ago, I learnt (and later taught) the opposite, but maybe we define "portable" differently and / or things may have changed in the last 25 years.  While `printf` is much easier to apply for C-programmers, IIRC `echo` was the way to go since / in pre-POSIX times to be portable between UNIXes and various shells (sh, csh, tcsh etc.).

> >  using string comparisons (per test command / built-in) is much faster than numerical comparisons.
> 
> Just out of curiosity, do you have a source for this? Because usually it's the other way around...

Oh, another bold statement, which made me ask "Just out of curiosity, do you have a source for this?"! :wink:
Yes, my own testing and shell programming guides (both a quarter of a century ago).  The core argument was, that shells (before bash) only know a single variable type: Strings.  Any other operation or evaluation (e.g., numerical ones) has to convert the variables involved from and to strings before and after it.  I remember having seen speed differences greater than factor 2.

But as an engineer, I prefer practical and relevant assessments (instead of theorising) and compared
`a=0; ns=$(date +%s%N); while [ $a != 100000 ]; do a=$(( $a + 1 )); done; echo $(( $(date +%s%N) - $ns ))` with
`a=0; ns=$(date +%s%N); while [ $a -ne 100000 ]; do a=$(( $a + 1 )); done; echo $(( $(date +%s%N) - $ns ))` under dash, bash and (accidentally) bash in sh-mode (on CentOS 8.5).
First insight: You are not right.  But me also not really.  For `dash` and bash in `sh`-mode the string comparison is faster by 10 - 15 %, but the variance of the results is of the same magnitude.  For bash called as `bash` it does not seem to make a difference, but the variance is even higher.
Second insight: `dash` is much faster than `bash`, with about 2/5 (40%) of the run-time. 
Side note: The biggest speed-up compared to writing classic, portable shell scripts comes from arithmetic evaluation per `$(( ))`, instead of wrangling with the historic, shell-external `expr` command.  Though I still have to research, if that is defined by POSIX for the Bourne shell (and would be happy if it is).

Olf0:
> #### 3.
> 
> Hence the only optimisation left to address is
> 
> > > Consider pulling `mkdir -p` out of the for loop.
> 
> I would be glad, if you review that similarly quickly when I get around to create another MR.

No need to, because I decided to discard that as well: The current `mkdir -p "$PM_PATCH_BACKUP_DIR"` is shorter, likely faster and easier to understand than the equivalent `if [ ! -d "$PM_PATCH_BACKUP_DIR" ]; then mkdir "$PM_PATCH_BACKUP_DIR"; fi` (the directory above always exists: `[PM_PATCH_BACKUP_DIR="$PM_PATCH_BACKUP_ROOT_DIR/$1"](https://github.com/sailfishos-patches/patchmanager/blob/master/src/tools/pm_apply#L16)`.

<br />
But thoroughly re-reading the new mangle functionality in the `pm_apply` script made me creating three new MRs (#216, #217 and #219) plus issue report #218: Please review.

nephros:
> #### 1.
> > `printf` is much, much more portable then `echo`. If anything, echo should be removed from any and all scripts out there.
> 
> Uh, this is a bold statement. Once upon long ago, I learnt (and later taught) the opposite, but maybe we define "portable" differently and / or things may have changed in the last 25 years. While `printf` is much easier to apply for C-programmers, IIRC `echo` was the way to go since / in pre-POSIX times to be portable between UNIXes and various shells (sh, csh, tcsh etc.).

We seem indeed ot have been taught different things - POSIX or not, `echo` has never been portable *if* called with any arguments apart from the string to output (like the usual `-n` `-e` and friends), as different implementations of shell builtins and `/bin/echo` had/have differing syntax or do not support options at all.

`printf`, while kind of un-shell-like, has always been defined  a bit better through the C language, and avoids the problem of supporting certain options and their ordering.

nephros:
> #### 2.
> > > using string comparisons (per test command / built-in) is much faster than numerical comparisons.
> > 
> > 
> > Just out of curiosity, do you have a source for this? Because usually it's the other way around...
> 
> Oh, another bold statement, which made me ask "Just out of curiosity, do you have a source for this?"! wink Yes I do, my own testing and shell programming guides (both a quarter of a century ago). The core argument was, that shells (before bash) only know a single variable type: Strings. Any other operation or evaluation (e.g., numerical ones) has to convert the variables involved from and to strings before and after it. I remember having seen speed differences greater than factor 2.

Good points. My "usually it's the other way around" stems more from the fact that in most (typed) programming languages integers are primary/primitive types which are easy/cheap to compare. And strings get very complicated to deal with very quickly as soon as you leave a plain old 7-bit ASCII "`char[]`", what with encodings, byte/character lengths, localizations etc.

Soo, basically the opposite of what you describe in the latter paragraph, and soething that as you say does not apply to shells  (at least not on the user-facing side of things).

Olf0:
> > #### 1.
> > > `printf` is much, much more portable then `echo`. If anything, echo should be removed from any and all scripts out there.
> > 
> > Uh, this is a bold statement.
> 
> We seem indeed ot have been taught different things - POSIX or not, `echo` has never been portable _if_ called with any arguments apart from the string to output (like the usual `-n` `-e` and friends), as different implementations of shell builtins and `/bin/echo` had/have differing syntax or do not support options at all.

Thanks for pointing this out.  With these hints, I was able to find quite sone references, e.g.:
* https://unix.stackexchange.com/a/565793
* https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo

> `printf`, while kind of un-shell-like, has been defined a bit better through the C language, and avoids the problem of supporting certain options and their ordering. (Of course, there certainly are non-portable variants of the `printf(1)` `format` string but I'll ignore that for now.)

Yes, it seems to be less evil (better portable) across shells and UNIXes.
Still some pitfalls may occur, see e.g., https://unix.stackexchange.com/questions/558636/bash-printf-formating-not-working

> > #### 2.

Simple shells seem to really handle every variable as a string internally.
I am quite sure `bash` does not, as most modern "un-typed" languages internally assign types to variables.
<br />

P.S.: Thank you, your comments sent me on a little research excursion for some hours, but that was well worth it: I learnt something and freed myself from some baseless belief.

PR #217, opened 2021-12-21
[pm_apply] Fix manglelist path

Olf0:
Use absolute paths for testing *and* sourcing.


Olf0:
IMO the same path must be tested, from which the content is `source`ed from subsequently.
I.e. here, use absolute paths for testing *and* sourcing.

nephros (approved):

Issue #218, opened 2021-12-21
Fix disabling of mangling?


Olf0:
While reviewing the shell script `pm_apply`, specifically the new functionality of mangling library paths, I wondered how the disabling of the whole mangling is supposed to work.

a. In [line 27](https://github.com/sailfishos-patches/patchmanager/blob/master/src/tools/pm_apply#L27) the variable `DISABLE_MANGLING` is evaluated for being zero length, but never set (at any place I am aware of); hence I assume it is meant to be set as an environment variable.
  **Q**: Is this documented somewhere?

b. If `DISABLE_MANGLING` contains something or sourcing the file `/etc/patchmanager/manglelist.conf` does not set the variable `MANGLE_CANDIDATES` to something, this variable stays empty (i.e., initialised, but zero length).
   But then the `for` loop in the function `mangle_libpath` is entered once with an empty string for `p`, consequently `totl_lines` = `cand_lines` = n (with n ≥ 1), … IMO this cannot yield anything right.
   
   Hence I suggest to prefix [the call of the function `mangle_libpath`](https://github.com/sailfishos-patches/patchmanager/blob/master/src/tools/pm_apply#L237) (in the script body) with
   `[ -n "$MANGLE_CANDIDATES" ] && ` per MR #219.

nephros:
Thanks or review, analysis and fixing MR, merged.

Which leaves open the nonexisting documentation - what would be the best place to put that? As far as I'm aware we don't have an "innards of PM" section of documentation at the moment.

Olf0:
> Which leaves open the non-existing documentation - what would be the best place to put that?

The [PM-wiki here at GH](https://github.com/sailfishos-patches/patchmanager/wiki).

> As far as I'm aware we don't have an "innards of PM" section of documentation at the moment.

Correct.  New sections are easy to create, many of the extant ones are just stubs, others contain some collected information which needs to be overhauled.  Thus creating proper content is the hard but crucial part.
Feel free to draft a new section; I may try as well but have enough open tasks, that this will be mid-January, earliest.  At least I now know what to document; thanks for that information.

Olf0:
But we do already have a section for environment variables, see https://github.com/sailfishos-patches/patchmanager/wiki/Environment-variables
IMO it fits well there.

nephros:
Documented.

[...]

PR #219, opened 2021-12-21
[pm_apply] Fix case of empty MANGLE_CANDIDATES

Olf0:
Fixes issue #218 b.


Olf0:
Oh, it obviously I was on the wrong branch (the one which includes the single commit of MR #217), when adding the single commit of this MR.
Actually I am too lazy to fix this, because both are trivial and should be consensual, so I will simply merge MR #217 before MR #219 and all will be fine. :slightly_smiling_face:

nephros (approved):

PR #223, opened 2022-01-04
Do convert multiple libpath references

Olf0:
Re-do commit b1fea73


Olf0:
FAIL, because I overlooked a flaw twice.
… re-do it once again!

PR #237, opened 2022-01-15
pm_apply: improve regexp

nephros:
corrects regular expressions:

1. '+' has special meaning, escape it
2. handle patches with "git-style" prefixes in the diff header, see [Test Cases #232 (comment)](https://github.com/sailfishos-patches/patchmanager/discussions/232#discussioncomment-1966155)

### DIFF
--- a/src/tools/pm_apply
+++ b/src/tools/pm_apply
@@ -102,8 +102,8 @@ mangle_libpath() {
       fi
 
       for p in $candidates; do
-        totl_lines=$(grep -c "^+++" "$PATCH_PATH")
-        cand_lines=$(grep -c "^+++ $p" "$PATCH_PATH")
+        totl_lines=$(grep -c "^\+\+\+" "$PATCH_PATH")
+        cand_lines=$(grep -c "^\+\+\+ [^/]*$p" "$PATCH_PATH")
         if [ $cand_lines -eq 0 ]; then
           continue  # nothing found, try next
         else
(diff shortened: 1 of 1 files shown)


Olf0:
Notes:
1. Ad "1. '+' has special meaning, escape it":
    AFAIK only for PCREs (Pearl complex REs) and maybe for the obsolete EREs, but `sed` is special so often and I regularly become confused between the various RE-styles: It can do no harm to escape each `+`.
2. Ad "handle patches with "git-style" prefixes in the diff header":
    AFAIU from the diff documentation, these are also simply path elements, but of relative paths.  See [GNU Diffutils: Detailed Description of Unified Format](https://www.gnu.org/software/diffutils/manual/diffutils.html#Detailed-Unified) and especially [GNU Diffutils: Avoiding Common Mistakes](https://www.gnu.org/software/diffutils/manual/diffutils.html#Avoiding-Common-Mistakes).
    I intended to argue that the character set for the first path element should be much more limited than `[^/]` (e.g., to `[[:alnum:]]`), but then realised that it could even contain backslash-quoted spaces ("`\ `").  So I have no better idea how to handle this.

Olf0 (approved):
Looking O.K.

nephros:
> I intended to argue that the character set for the first path element should be much more limited than [^/] (e.g., to [[:alnum:]]), but then realised that it could even contain backslash-quoted spaces ("\ "). So I have no better idea how to handle this.

Yeah I was reluctant to use such broad matching. Correct one would be "any character allowed in a linux file path, including illegal characters if escaped properly". But even if such a regex term exists it would probably just look like line noise, hence not very maintainable.  
So I opted for the easily readable might-break-in-some-corner-cases option.

PR #238, opened 2022-01-15
remove superfluous and confusing log output

nephros:
These output meaningless things to the log file.

PR #239, opened 2022-01-16
(Re-)move misplaced log output and counter initialisation

Olf0:
This enhances PR #238, while retaining its intention.

### DIFF
--- a/src/tools/pm_apply
+++ b/src/tools/pm_apply
@@ -46,11 +46,6 @@ log() {
   echo "$@" | tee -a "$PM_LOG_FILE"
 }
 
-log "Mangle candidates"
-log $MANGLE_CANDIDATES
-log "Disable mangling"
-log $DISABLE_MANGLING
-
 failure() {
   log
   log "*** FAILED ***"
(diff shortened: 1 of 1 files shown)


Olf0:
Thank you for making me look at `pm_apply`, again.
I would prefer to output the final mangle-list, as in PR #239 (in contrast to your suggestion PR #238), because this is what ultimately is used.
But I fully agree with the fundamental observation which resulted in PR #238: Logging the mangle list as debug-output, even when it is not used, is confusing and superfluous.  As for the `DISABLE_MANGLING` variable, which results in an empty mangle list (which is obvious).  And the original output format, which makes empty variables not obvious. 

Plus: Each time I look at this, I detect some other flaws:
* Move the initialisation of the `found` variable where it belongs: Right before the loop.
* Remove a superfluous space character.
* Enhance a comment.

Are you O.K. with merging PR #239 to the `Olf0-patch-3` branch and ultimately merge that into `master`?
And to close PR #238 unmerged.

nephros (approved):

PR #240, opened 2022-01-18
l18n: Adapt to bitness nomenclature and fix all line number references

Olf0:
…, see https://github.com/sailfishos-patches/patchmanager/wiki/Terms-and-wording#0---9

And check all and adapt all but one line numbers in translations/settings-patchmanager.ts, including …
fix quite some line number references being broken for long.

### DIFF
--- a/src/bin/dialog/dialog.qml
+++ b/src/bin/dialog/dialog.qml
@@ -60,8 +60,8 @@ ApplicationWindow {
         Page {
             onStatusChanged: {
                 if (status == PageStatus.Active && !appWindow.remorseItem) {
-                    remorse.execute(button, qsTranslate("", "Activate all Patches marked active"), function() {
-                        console.info("Accepted activation of all Patches marked active.")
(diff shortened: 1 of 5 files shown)


nephros (approved):
Fine!

About the line numbers: usually they should be updated automatically by `lupdate`, no need to handle them manually if that is what happened here.

They do become "out of date" of course by merging things and not doing a `lupdate` run after.

Olf0:
> About the line numbers: usually they should be updated automatically by `lupdate`, no need to handle them manually if that is what happened here.

Yes, this is what I did, because without using the SFOS-SDK and Qt-Linguist, even no local files (i.e., no local git repository), there is no other way.  I am not keen on deploying all this local infrastructure.

> They do become "out of date" of course by merging things and not doing a `lupdate` run after.

AFAIU, `lupdate` also updates the strings proper: https://doc.qt.io/qt-5/linguist-manager.html#using-lupdate
Thus can you please run it to determine if I missed something; if so, please create a PR with the `translations/settings-patchmanager.ts` file it created.

Issue #220, opened 2021-12-21
Optimise manglelist handling (Was: Fails if a Patch contains multiple sections matching different elements on the manglelist!)


Olf0:
AFAICS, the function `mangle_libpath` fails if a Patch contains multiple sections matching different elements on the [manglelist.conf](https://github.com/sailfishos-patches/patchmanager/blob/master/src/etc/manglelist.conf#L9) (which is [`source`d as `MANGLE_CANDIDATES` in `pm_apply`](https://github.com/sailfishos-patches/patchmanager/blob/master/src/tools/pm_apply#L28)):
The `return` statement is placed right before the `done`, hence the `for` loop is run once at most (i.e., it never loops).

But AFAICS, a Patch might deliberately contain multiple sections matching different elements on the [manglelist.conf](https://github.com/sailfishos-patches/patchmanager/blob/master/src/etc/manglelist.conf#L9), thus the `for` loop should iterate through the elements of `MANGLE_CANDIDATES`.

Am I missing some aspect, or is this really a flaw worth pondering about (i.e., developing an enhancement)?

nephros:
I agree that return statement should go.  
I'll have to set up a proper test case before I can say for sure whether that's a one-liner or more changes are required. I don´t trust my brain with analyzing loops.

Olf0:
> yeah, no, we need some more involved re-looping if we want to support "mixed" patches, i.e. patches which match more than one "candidate".

I will ponder about it and likely come up with a suggestion.
It should not be too hard, rather getting this loop closer to general programming guidelines (i.e., "do avoid jumping out of loops, rather adapt the loop for that to become unnecessary"), which makes it simpler to comprehend.

P.S.: As this is shell code, I should be able to handle that well.  What I am rather concerned of, after finding a couple of flaws in these few lines of new code, are the "mangling"-related changes elsewhere, which are way harder for me to understand due to being written in languages foreign to me.  I know you partially or mostly authored them, which makes reviewing hard, but maybe after many weeks you can look at these changes / additions as if they were written by someone else; i.e., can you try to review MRs #109 and #140 again (or better, but I fail to find it: whichever MR merged the "patch-bitness" branch).

Olf0:
<sup>Missed to mention this:</sup>
Yes, please do create a test case, sometime (not urgent, i.e., take your time).

Olf0:
[...]

PR #241, opened 2022-01-18
[pm_apply] improve mangling by patching outside the loop

nephros:
See also #220.

This moves the sed call outside of the detection/mangle list loop.
sed command is built in the loop, but only one call to sed is made.

This saves a couple of greps and tests as well.

Tested against Test Case 2-2, from https://github.com/sailfishos-patches/patchmanager/commit/9fb6adb67a4a0948eeb731c08fc2dfa17a88f6c3#diff-bdc16c81678138d1bc10b9ac315b66564aaf85cad91efaed15e375ef837fd7a3

### DIFF
--- a/src/tools/pm_apply
+++ b/src/tools/pm_apply
@@ -99,42 +99,47 @@ mangle_libpath() {
 
     found=0
     sedcmd=""
-    # look for lines to replace, if found add pattern to sed commad
+    # look for lines to convert, if some are found add pattern to sed scriptlet
     for p in $candidates; do
       cand_lines=$(grep -c "^\+\+\+ [^/]*$p" "$PATCH_PATH")
       if [ $cand_lines -eq 0 ]; then
         continue  # nothing found, try next
-      else
-        found=$(( $found + $cand_lines ))
       fi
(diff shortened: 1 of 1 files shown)


Olf0 (approved):
Pretty much what I had in mind (thanks!), except for using sed "streamingly", so the old file is only read once and the new one only written once.
Plus some beautifying.

Please re-review.

nephros:
> Please re-review.

Done, Sweet. 

I can't click that button to review because I am the PR author, so I'll just hit merge :)

Olf0:
> I can't click that button to review because I am the PR author, 

Yes, that is idiotic logic by the GitHub coders: I was also unable to request a review from you.

> so I'll just hit merge :)

Which is fine, because you surely are going to test that once more before the 3.2.1 release, right? :wink:

nephros:
> Which is fine, because you surely are going to test that once more before the 3.2.1 release, right? 😉

Always.
I only have one 32bit device to test on though.

b100dian:
On it - so I test this PRs build?

b100dian:
Works on my machine (return old pulley menu) on 64 bit

PR #248, opened 2022-01-23
update translation source

nephros:


Olf0:
Thank you very much.  <sub>While I checked that the brackets "**(**" and "**)**" do not have to be HTML-encoded, I forgot about the apostrophe "**'**"→"\&apos;": Really not a good idea to perform such changes manually, but "unexpectedly" (as so often :wink:) a task originally assumed to be small and simple "ballooned", here into the "big strings rework".</sub>

Note, that I reverted one change, because it indicated that I missed to adapt a source string: Hence I rather fixed that source string and added this commit to this branch.

Olf0 (approved):
Looking good.

Olf0:
@nephros, contrary to our usual process, I will merge this now, so I can check tonight at Transifex, that this is looking good now.

PR #258, opened 2022-01-27
Add Test Cases for patching

nephros:
See https://github.com/sailfishos-patches/patchmanager/discussions/232

This should probably *not* be squashed on merge so the individual cases keep their histories.


Olf0:
@b100dian, IMO this is not at all critical WRT time, so take all the time you need to review this comfortably.  But I requested your review (rather meant as "suggested to review this to you"), because I feel a review from me will be far less stringent due to far less know-how (and I already looked over this PR without seeing anything I could criticise).

b100dian:
Thanks for adding much missed test cases!
@nephros can you add some details how this works? (in the description of the PR would do)
Specifically I don't understand when the patches are applied.

What I tried: I started the app from the CI build and it failed with 
```
$ /usr/bin/sailfish-qml patchmanager-testcase
[W] unknown:3 - file:///usr/share/patchmanager-testcase/qml/patchmanager-testcase.qml:3:1: module "org.SfietKonstantin.patchmanagertests" is not installed 
     import org.SfietKonstantin.patchmanagertests 1.0 
     ^
```
Then the app from debug deploy from Sailfish IDE -> starts, displays "Case 1 *unpatched* Case2 *unpatched*` (as expected by reading `patchmanager-testcase.qml`

nephros:
> Thanks for adding much missed test cases! @nephros can you add some details how this works? (in the description of the PR would do) Specifically I don't understand when the patches are applied.
> 
> What I tried: I started the app from the CI build and it failed with
> 
> ```
> $ /usr/bin/sailfish-qml patchmanager-testcase
> [W] unknown:3 - file:///usr/share/patchmanager-testcase/qml/patchmanager-testcase.qml:3:1: module "org.SfietKonstantin.patchmanagertests" is not installed 
>      import org.SfietKonstantin.patchmanagertests 1.0 
>      ^
> ```
> 
> Then the app from debug deploy from Sailfish IDE -> starts, displays "Case 1 **unpatched** Case2 **unpatched**" (as expected by reading `patchmanager-testcase.qml`

I can not reproduce that here, with the package installed (not using an IDE).

Having the `qmldir` file at $LIBDIR/fietKonstantin/patchmanagertests/qmldir should make that `import` statement work (even though there is no real QML module/plugin present) AFAIK.

On my device that looks like this:

```
nemo@PGXperia10:~/git/patchmanager/repo $ QML_IMPORT_TRACE=true qmlscene /usr/share/patchmanager-testcase/qml/patchmanager-testcase.qml  2>&1 | grep ertests
18:33:33.927 unknown:0 unknown QQmlImports(file:///usr/share/patchmanager-testcase/qml/patchmanager-testcase.qml)::addLibraryImport: "org.SfietKonstantin.patchmanagertests" 1.0 as ""
18:33:33.928 unknown:0 unknown QQmlImports(file:///usr/share/patchmanager-testcase/qml/patchmanager-testcase.qml)::importExtension: loaded "/usr/lib/qt5/qml/org/SfietKonstantin/patchmanagertests/qmldir"
18:33:34.277 unknown:0 unknown QQmlImports(file:///usr/share/patchmanager-testcase/qml/patchmanager-testcase.qml)::resolveType: "TestCase1Item" => "" QUrl("file:///usr/lib/qt5/qml/org/SfietKonstantin/patchmanagertests/TestCase1Item.qml") TYPE/URL
```

Why did I invent `org.SfietKonstantin.patchmanagertests`?
In order to have something that lives under LIBDIR so the 32/64 things can be tested.

So the test infrastructure consists of:

 1. The QML module under `LIBDIR/qt5/qml/...`, which is some QML component files and a `qmldir` file so they can be used as a module
 2. The little app as a QML file and .desktop file under `/usr/share/`, which imports the above module and thus shows changes to the module's components.
 3.  a set of patches which act on those files in various ways

Olf0:
Dear @b100dian, this seems to be a bit deadlocked between you both.  I am afraid that @nephros will have to regularly update this branch from the master branch (as it happened by [commit 67f8b64](https://github.com/sailfishos-patches/patchmanager/pull/258/commits/67f8b6406d9d58d52277bbe3664923ff859d25bb)), until this is merged.

I considered a "best effort review" from me, which will only scratch the surface because this contains too much non-trivial C++ and QML code for me to review properly.  But as @nephros is successfully using the test cases to generate meaningful results and I am happy that he does, I trust that this is working.  At least I could let this stalled PR progress this way.

OTOH I value your knowledge and practical programming experience very much @b100dian, so I would rather like to see this properly reviewed (i.e., by you, not me).  But if you do lack the time in the foreseeable future, just state so and we deviate from the standard procedures: I perform a "best effort review", this is merged and your review is postponed; although I see the risk of this becoming postponed to "never day" as a consequence.

All in all this is not altering existing code, and not adding code, which is deployed to users, AFAICS: Hence I do not believe anything really bad can happen ("famous last words™").

Olf0:
P.S.: *»I'll cop out of this one and refer to "read the source Luke" here.«* made me chuckle.  "But I also want nicely written, in-depth documentation for the test cases in our wiki"!  While that is true, I rather have test cases without much documentation than none.  Even though I am afraid of the day on which I will face the "read the source Luke" part in practice.

More on the serious side:
IMO the prime objective currently is to shove the test-cases into the master branch while ensuring that they are basically working and usable, in order to get them utilised by more people than @nephros alone.
So yes, if there are issues with the test-cases beyond *"What do the patches do?"* I think we should clarify these and address them first.  And it sounds as if @b100dian noticed that one may run into such issues, assuming that "Sailfish IDE" addresses the "SailfishOS SDK / emulator", which presumably many use for development (not me, though).

As @nephros addressed the "`import` issue", so this should be at least manually replay-able under the SFOS-SDK, I have to admit that I do not really understand the remaining open question *"Specifically I don't understand* when *the patches are applied."*  In complete ignorance (i.e., not having tested any of this yet) I would assume, when one activates them in PM GUI.  An further assume (with further ignorance) that the testcase app should then display displays "Case 1 **patched** Case2 **patched**", when started again.

nephros:
> In complete ignorance (i.e., not having tested any of this yet) I would assume, when one activates them in PM GUI. And further assume (with further ignorance) that the testcase app should then display "Case 1 **patched** Case 2 **patched**", when started again via its desktop icon or `sailfish-qml patchmanager-testcase` with `$LIBDIR/SfietKonstantin/patchmanagertests/qmldir` present.

Exactly, sorry for not making that clear in my reply.

So the answer to  "when are patches applied" is "when you apply them" i.e. using the PM GUI as usual, or `/usr/sbin/patchmanager -a <patchname>`.

b100dian (approved):
Thank you both - not intending to deadlock this at all:)
The simple explanation for how to test the patches eluded me, I was looking for some nonexistent automated process and I'm actually glad it is as simple as the added Readme points out :relieved: 
For the error I've seen - I don't have problems with building the test package myself. 

The problem with the import is just with the output_rpms attached to the [checks](https://github.com/sailfishos-patches/patchmanager/pull/258/checks) of this PR and possibly what will come out in say, chum/OBS build. I haven't had that problem when building myself.

nephros:
OK thanks all.

I will squash/rebase a couple of these, and then merge this.

PR #259, opened 2022-01-30
Record time to apply all patches in startup dialog

nephros:


Olf0:
1. I suppose `%1` contains "seconds" in its value.  If not, please add the word "seconds".
2. If I understood you correctly, one needs to run `lupdate` after touching any file with translated strings (as soon as line numbers are skewed), which updates the primary `ts` file: These updates seem to be missing.
3. Cannot judge about the added `;`.

nephros (commented):
This change in time() / timeEnd() will result in an output in the log file of:

`8:47:05.783 unknown:0 unknown autoApplyingRuntime: 1231ms` as opposed to a sentence. Not sure whether that was the intention. I'm fine with either, it's just a console log message.

nephros:
> 1. I suppose `%1` contains "seconds" in its value.  If not, please add the word "seconds".

The Formatter will return NN:MM with either mm:ss or hh:mm depending on time. So I guess "minutes"could be added but is not correct in case it takes more than an hour (unlikely).

Olf0:
> This change in time() / timeEnd() will result in an output in the log file of:
> 
> `8:47:05.783 unknown:0 unknown autoApplyingRuntime: 1231ms` as opposed to a sentence. Not sure whether that was the intention.

Not really, but …
> I'm fine with either, it's just a console log message.

so am I: For a log message it is fine (although now I see, why the ID was also a sentence).
My main objective was to let the ID be a classic ID.

What I am more concerned about, because it is an user facing message, is then time string formatter: Can't you simply output the value as full or fractional seconds or full milliseconds (no matter how many, e.g. 12345), without any complications (i.e., avoiding some XX:YY format)?   IMO realistic runtimes are between 0,2 and 30 seconds.  I would still prefer full seconds to keep it simple and easy to comprehend, but fractional seconds or full milliseconds are sure fine, too.

nephros:
> IMO realistic runtimes are between 0,2 and 30 seconds. I would still prefer full seconds to keep it simple and easy to comprehend, but fractional seconds or full milliseconds are sure fine, too.

That's trivial, simply remove any formatting (and introduce fancy localized number display through `%L1` arguments instead).

Olf0:
> > IMO realistic runtimes are between 0,2 and 30 seconds. I would still prefer full seconds to keep it simple and easy to comprehend, but fractional seconds or full milliseconds are sure fine, too.
> 
> That's trivial, simply remove any formatting (and introduce fancy localized number display through `%L1` arguments instead).

Oh, this is nice.  I also like [the examples in the Qt-5 documentation for `%Lx`](https://doc.qt.io/qt-5/qtquick-internationalization.html#5-use-lx-so-numbers-are-localized), which show two fractional digits, because displaying milliseconds or (equivalently) seconds with three fractional digits is of too finely granular, while rounded seconds result in "0" for anything less than 0,5 seconds.
So, If I understood correctly that times are measured in milliseconds (plus the syntax), then altering [line 176](https://github.com/sailfishos-patches/patchmanager/pull/259/files#diff-61a9d678f8b1c8eb204e1691ff0b7f5db8abb7efe152a2f9b546809d75cc412aR176) to
`var runtime = Math.round( ( t - progress.runTimeStart ) / 10 ) / 100 ;`
might achieve that, right?

P.S.: BTW, what is the background of introducing all the semicolons? I believed they are superfluous at line ends for JavaScript (as in shell scripts), but always avoided JS as much as I could.

Olf0:
Side note: It believe we shall not merge this before the v3.2.1 release, because it introduces strings changes, which trigger a translation another round.

nephros:
> Side note: It believe we shall not merge this before the v3.2.1 release, because it introduces strings changes, which trigger another translation round.

ACK, it's not terribly urgent.

nephros:
> P.S.: BTW, what is the background of introducing all the semicolons? I believed they are superfluous at line ends for JavaScript (as in shell scripts), but always avoided (to learn, study, comprehend) JS as much as I could.

This might be different in current iterations of the standard, but the original JS engines had semicolons to end statements, but *automatically inserted* them where it thought they were missing,. Which obviously can lead to very unexpected effects. 
(Similar in a way to the "tag soup" way of interpreting HTML in the early days.) 

There's some somewhat entertaining "why javascript sucks" rants out there about this.

So tihs is the background of the change, while I have no idea whether it makes any difference in QML. Likely not.

Olf0:
> > Side note: It believe we shall not merge this before the v3.2.1 release, because it introduces strings changes, which trigger another translation round.
> 
> ACK, it's not terribly urgent.

OTOH, a release can be made any time.  The criterion is rather, if we expect any more translation work from Transifex.  Except for this, v3.2.1 is ready IMO.
BTW, can you please (quickly) review the German translation, which I completed a few weeks ago.

Olf0:
> So, If I understood correctly that times are measured in milliseconds (plus the syntax), then altering [line 176](https://github.com/sailfishos-patches/patchmanager/pull/259/files#diff-61a9d678f8b1c8eb204e1691ff0b7f5db8abb7efe152a2f9b546809d75cc412aR176) to `var runtime = Math.round( ( t - progress.runTimeStart ) / 10 ) / 100 ;` might achieve that, right?

I just pushed commit [c64f65a](https://github.com/sailfishos-patches/patchmanager/pull/259/commits/c64f65a82a9fd23ba74cbf964ac735c6b691d829), hopefully it is syntactically correct: Please review.

nephros:
> So, If I understood correctly that times are measured in milliseconds (plus the syntax), then altering [line 176](https://github.com/sailfishos-patches/patchmanager/pull/259/files#diff-61a9d678f8b1c8eb204e1691ff0b7f5db8abb7efe152a2f9b546809d75cc412aR176) to
> `var runtime = Math.round( ( t - progress.runTimeStart ) / 10 ) / 100 ;`
> might achieve that, right?

Yes, correct
.
Alternatively one could just check that at display time and say

                    label.text = runtime < 1 
                         ? qsTranslate("", "Activating all enabled Patches took less than a second." )
                         : qsTranslate("", "Activating all enabled Patches took %L1 seconds.").arg(runtime);

Olf0:
> There's some somewhat entertaining "why javascript sucks" rants out there about this.

I have no idea about all this, because 20 years ago I concluded that "javascript sucks" and stayed away from it.

> So tihs is the background of the change, while I have no idea whether it makes any difference in QML. Likely not.

Well, the [QT5-demos](https://doc.qt.io/qt-5/qtquick-codesamples.html) use them, but much more sparingly (maybe also: inconsistently?): AFAIU only after certain types of expressions, e.g., setting variables per `=`.

I wonder what the current rule / recent guidance is, and how to proceed: There are many more QML files in Patchmanager's source tree (without any "`;`" AFAICS).

Olf0:
> and have those who need more precise information to read the console log which has the milliseconds.

Nerds love to see how ultrafast their device is: Thus "0,15 seconds" is a great output! :wink:

Olf0 (approved):
Because all the replies to my checks & balances questions sound very reasonable and hence I believe to understand the changes, I am approving this despite not being a QML / JavaScript coder.

Though I still have the impression that the [mentioned QT5-demos](https://doc.qt.io/qt-5/qtquick-codesamples.html) use "`;`" less often, basically only when assigning values to variables.
But as it is just a few "`;`" and they were almost completely missing before, this clearly constitutes an improvement (and I do trust your opinion as a programmer, especially for languages I do not use).

PR #271, opened 2022-02-20
DeveloperPage: Remove AUSMT entry

Olf0:
… because AUSMT is not used by Patchmanager any more for long (since v3-alpha).

### DIFF
--- a/src/qml/DevelopersPage.qml
+++ b/src/qml/DevelopersPage.qml
@@ -43,16 +43,6 @@ Page {
     ListModel {
         id: model
         ListElement {
-            icon: "/usr/share/patchmanager/data/webosinternals.png"
-            category: "AUSMT"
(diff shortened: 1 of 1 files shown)


Olf0:
While I love to document history in order to mare it traceable, a better web-link as a reference for AUSMT would have been https://www.webos-internals.org/wiki/Auto-Update_System_Modification_Technology

But very unfortunately most FLOSS developments around WebOS have stalled for long.
For example, note the certificate error by using an outdated Let's Encrypt certificate at https://webos-internals.org or when various web-pages in their wiki were last updated.

Anyway, Patchmanager does not use AUSMT since [v3-beta in 2018](https://talk.maemo.org/showthread.php?p=1544926#post1544926) any more.

nephros (approved):

Issue #20, opened 2017-08-13
[Usability issue] Please move the entry "Unapply all patches" in Patchmanager's top level pulley menu one or two positions further up, i.e. *above* the "About" entry


Olf0:
**Background:** I accidentally selected "Unapply all patches" numerous times when trying to enter Patchmanager2's web catalog. 
As entering the web catalog is often needed, the pulley menu entry "Unapply all patches" should be moved further away from the web catalog's entry.
Hence please move the entry "Unapply all patches" in Patchmanager2's top level pulley menu one or two positions further up, i.e. *above* the "About" entry. 

Tested with Patchmanager 2.3.1-10.39.1.jolla on a Jolla 1 phone under SFOS 2.1.0.11

Originally reported [on TMO as issue 11.](http://talk.maemo.org/showpost.php?p=1529100&postcount=2249) (filing this here [as promised](http://talk.maemo.org/showpost.php?p=1529055&postcount=2245)).

nephros:
Well, fat-fingering the Pulley menu is something that happens, almost no matter where the menu options are placed.

As a countermeasure, how about adding a Remorse popup?  
That's the usual way of SFOS dealing with destructive changes, is easy to implement and avoids the [xkcd-1172](https://xkcd.com/1172/) effect that a reordering of the menu would have.

Olf0:
O.K., also an apropiate solution (I assumed reordering to be less intrusive / less work).

PR #272, opened 2022-02-20
PatchmanagerPage: Reorder pulley menu entries

Olf0:
… "Settings" shall be below "About" and easier to reach, plus "Deactivate all Patches" the hardest to reach (i.e., topmost)

### DIFF
--- a/src/qml/PatchManagerPage.qml
+++ b/src/qml/PatchManagerPage.qml
@@ -139,8 +139,9 @@ Page {
             busy: view.busy
             enabled: !busy
             MenuItem {
-                text: qsTranslate("", "Settings")
-                onClicked: pageStack.push(Qt.resolvedUrl("SettingsPage.qml"))
+                text: qsTranslate("", "Deactivate all Patches")
+                onClicked: menuRemorse.execute( text, function() { PatchManager.call(PatchManager.unapplyAllPatches()) } )
+                visible: PatchManager.loaded
             }
 
             MenuItem {
@@ -149,9 +150,8 @@ Page {
             }
 
             MenuItem {
-                text: qsTranslate("", "Deactivate all Patches")
-                onClicked: menuRemorse.execute( text, function() { PatchManager.call(PatchManager.unapplyAllPatches()) } )
-                visible: PatchManager.loaded
+                text: qsTranslate("", "Settings")
+                onClicked: pageStack.push(Qt.resolvedUrl("SettingsPage.qml"))
             }
 
             MenuItem {
--- a/src/qml/PatchManagerPage.qml
+++ b/src/qml/PatchManagerPage.qml
@@ -138,11 +138,16 @@ Page {
         PullDownMenu {
             busy: view.busy
             enabled: !busy
+            
+            /*
+            Disabled due to discussion at https://github.com/sailfishos-patches/patchmanager/pull/272#issuecomment-1047685536
+            
             MenuItem {
                 text: qsTranslate("", "Deactivate all Patches")
                 onClicked: menuRemorse.execute( text, function() { PatchManager.call(PatchManager.unapplyAllPatches()) } )
                 visible: PatchManager.loaded
             }
+            */
 
             MenuItem {
                 text: qsTranslate("", "About Patchmanager")


Olf0:
I revisited this topic (literally, i.e., by looking at it closely in its current state) and the order just appears to be awkward: Why is the "Settings" entry above the "About" entry, which is much less often used; and why are they both above the dangerous "Deactivate all Patches" entry?
While we discussed that in issue #20, and I was willing to reconsider my former assessment in the light that "Deactivate all Patches" with a remorse timer is far less dangerous (if one realises quickly enough what is happening), I ultimately came to the conclusion that the current order just does not make any sense, as described above (i.e., it is more than just the position of the single "dangerous" entry).

Hence I "just did it": Reordered the topmost three entries so it does make sense (to me, at least).

nephros:
> … "Settings" shall be below "About" and easier to reach, plus "Deactivate all Patches" the hardest to reach (i.e., topmost)

I disagree with you there. In my experience, top menu entries are *easiest* to reach by acccident (maybe tied with the first/lowest entry when you somehow lose touch).
(They are also the easiest to reach on purpose, by vigurously flicking.)


Also, it's very common for in the SFOS app space to have About at the top, Settings next, and all others below those. We should continue to follow that layout.

Olf0:
> > … "Settings" shall be below "About" and easier to reach, plus "Deactivate all Patches" the hardest to reach (i.e., topmost)
> 
> I disagree with you there. In my experience, top menu entries are _easiest_ to reach by acccident (maybe tied with the first/lowest entry when you somehow lose touch). (They are also the easiest to reach on purpose, by vigurously flicking.)
> 
> About the general Pulley menu order, it's very common for in the SFOS app space to have About at the top, Settings next, and all others below those. We should follow that layout.

Yes!  Luckily we have converging opinions here (with issue #20 as a starting point), because I always perceived it as very awkward that Patchmanager has "Settings" as the topmost entry and "About" directly below it.

So the only point left where we do still not concur is the position of the "Deactivate all Patches" entry.
Some context, plus my consideration and experiences:
* The pulley menu entry "Deactivate all Patches" is a relict from Patchmanager **2**, because with PM 1 & 2 one must deactivate ("unapply" before PM 3.2.1) all Patches before upgrading SailfishOS or "hell will break loose"; seriously, SailfishOS installations were often broken beyond repair, because it was not feasible for users to determine which packages to reinstall to "heal" a specific installation: Factory-reset or flashing anew was the "easy way out", but a very tedious one.
* Since PM **3** this pulley menu entry has become superfluous.  This fuctionality is also available at the command line: `patchmanager --unapply-all`.  And I also experienced recently [how easily PM 3 enters "fail safe mode"](https://codeberg.org/nephros/install-history/issues/14#issuecomment-378961) (without even being guilty, but it sure cannot detect that), while having seen the code for "fail safe detection and mode" for the first time a few months ago.
   I believe it would be / have been a good maintainers task to eliminate it in PM ≥ 3.0.0, to save people from still performing this tedious back and forth of deactivating all Patches before upgrading SailfishOS and re-activating the non-conflicting ones after the upgrade ("because I have heard that this is necessary"), which has become a voodoo ceremony (i.e., black magic, which only works for the ones believing in it).
   The only thing I am unsure about how many people (and how forcibly) will complain, because they cannot perform their voodoo ceremony any more (this goes a bit in line with Peter's point depicted via XKCD in issue #20), and if it is better to endure that or to keep offering this pulley menu entry.  Personally I am fine with either way.
* In my experience I slipped (i.e., lost control) most often somewhere in the middle of long pulley menus (i.e., with a lot of entries): Specifically for Patchmanager I did accidentally select  menu entry "Unapply all Patches" (when it was still called that) at least three times, because it is nicely placed in the middle of a pulley menu with at least 4 entries.  Which sent me right into hell to figure out, which combination of non-conflicting Patches (out of then installed ones) I had activated before; one could easily spent an hour with that when using PM 2.  While this has become a lot easier to determine (possibly conflicting Patches are displayed in each Patch's details), I can tell from the recent "fail safe"-experiences that it still takes almost half an hour with >> 50 Patches installed (the first time; one becomes quicker when doing this repeatedly).
* You, Peter, have alleviated the whole issue by attaching a remorse timer to this action, *if* one is quick enough to realise what happened, and *if* one is not fooled by recently reversed logic of confirming the action (by a swipe or tap on the little cross, which formerly cancelled the action).
* No, "vigorously flicking" does not select the topmost entry: That lets the pulley menu fully stick out with some extra space atop the topmost entry and no entry selected.

Thus I am seriously asking either to "hide" this "dangerous" pulley menu entry where it does least harm (which is the topmost entry IMO) or to eliminate it (which likely is the proper solution).

nephros:
I have no objections to removing it competely, or add and option to show/hide it in the "Advanced" Settings.

Although I do find it useful sometimes and will problably patch it back in privately it it's removed ;)

About failure/panic mode:

I haven't looked into it, but I thing that is activated whenever lipstick crashes for any reason. It's hard for PM to detect whether who is to blame for that....  
I have also no idea yet what the "resolve failure" is supposed to do, it seems to do more or less nothing right now.

I would expect it to restore the configuration that was last saved*), of at least give to option to do that.

But these things are something to be discussed in a separate issue, not here.

*) flashbacks of Windows last good configuration

nephros:
I guess for this PR we have an agreement that reordering the Settings/About is what we'll do.

Please make a push doing that, and if you like comment out the "unapply all" Menu entry. If you do I will implement the Settings part on top later.

Olf0:
[Commit e140cfd](https://github.com/sailfishos-patches/patchmanager/pull/272/commits/e140cfded7178548f4386c22d81050ecd409588d) disables the "Deactivate all Patches" pulley menu entry.

Please check that I have done that correctly, I have to look up the f\*\*[king QML syntax](https://doc.qt.io/qt-5/qtqml-syntax-basics.html#comments) even for such simple things.
Side note: Did I already mention :stuck_out_tongue_winking_eye:: I hate JavaScript and QML :skull_and_crossbones:!  It was invented to enable those "web-developers" inflicting harm at other places than web-pages, which is easy to suppress in a controlled manner by NoScript, in contrast to QML.

nephros (approved):

PR #273, opened 2022-02-20
spec file: minor improvements

Olf0:


Olf0:
It is worse than expected: See issue #274.
Thus all comment lines have to be moved out of the `%description` section.

nephros (approved):

PR #276, opened 2022-02-22
remove binary from source tree

nephros:
a build artefact was introduced in 9431443 / #109


Olf0 (approved):
Oops: https://github.com/sailfishos-patches/patchmanager/commits/9431443c1fbb424d6aac05b67fb1a6cb2f8c5dc8/src/bin/patchmanager-daemon/patchmanager

Fully acked to delete it.

Issue #275, opened 2022-02-22
[BUG] Always shows "Failed" popup even if successful


nephros:
**SailfishOS VERSION** (Settings → About product → Build): 4.2
<br />**HARDWARE** (Settings → About product → Manufacturer & Product name): Xperia 10
<br />**PATCHMANAGER VERSION** (Settings → Patchmanager → [Top pulley] About): 3.2.1

#### BUG DESCRIPTION

When activating or deactivating patches, the "failed" popup is always shown, even if the action was successful

#### STEPS TO REPRODUCE

Activate or deactivate any patch

#### ADDITIONAL INFORMATION

The switch indicator activates or deactivates correctly.

nephros:
Putput from `devel-su -p jolla-settings` and deactivating an active patch:


```
11:29:53.418 unknown:0 unknown void PatchObject::unapply(QJSValue)
11:29:53.418 unknown:0 unknown void PatchObject::setBusy(bool) true
11:29:53.422 qml: file:///usr/lib/qt5/qml/org/SfietKonstantin/patchmanager/PatchManagerPage.qml:352 expression for onBusyChanged onBusyChanged: patch-email-plain true
11:29:53.425 unknown:0 unknown QDBusPendingCallWatcher* PatchManager::unapplyPatch(const QString&)
11:29:53.792 unknown:0 unknown void PatchManager::onToggleServicesChanged(bool) true
11:29:53.808 qml: file:///usr/lib/qt5/qml/org/SfietKonstantin/patchmanager/PatchManagerPage.qml:345 expression for onPatchedChanged onPatchedChanged: patch-email-plain false
11:29:53.808 unknown:0 unknown void PatchObject::setBusy(bool) false
11:29:53.809 qml: file:///usr/lib/qt5/qml/org/SfietKonstantin/patchmanager/PatchManagerPage.qml:352 expression for onBusyChanged onBusyChanged: patch-email-plain false
11:29:54.207 unknown:0 unknown void PatchManager::requestListPatches(const QString&, bool) "" false
11:29:54.241 unknown:0 unknown void PatchManagerModel::populateData(const QVariantList&, const QString&, bool) 37
11:29:54.241 unknown:0 unknown void PatchManagerModel::populateData(const QVariantList&, const QString&, bool) Altered: "" installed: false

```

nephros:
Debug journal log of a similar event:

```
[...]

PR #281, opened 2022-02-25
Fix no-notification logic

nephros:
Closes: #275


Olf0:
I would have assumed this to be equivalent (before and after the change), [apparently you thought so, too](https://github.com/sailfishos-patches/patchmanager/blame/61ea43c8bcc9d1b50924c5ed3214f3c9a7786f63/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L1889).
I only wonder, why [similar constructs](https://github.com/sailfishos-patches/patchmanager/blob/master/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L1376) at other places seem to work fine.

At least I understand now, why this was not in PM ≤ 3.2.0, only in 3.2.1: It was introduced 14 December 2022 and we made only that single release since then.

Olf0 (approved):

PR #282, opened 2022-02-25
Trivial: fix CR/LF line endings

nephros:


Olf0:
How did you realise it was?  Most things (editors etc.) hide that (too) well nowadays.

P.S..: So Andrey K. works under Windows, or at least he used to.

Olf0 (approved):

nephros:
> How did you realise it was? Most things (editors etc.) hide that (too) well nowadays.
> 

True. I changed something in the file, and `git diff` showed the `^M` CR symbol change for the edited lines ;)

PR #283, opened 2022-02-25
[UI] Allow "Unapply All" menu entry through settings

nephros:
Follow up on  #272, more precisely https://github.com/sailfishos-patches/patchmanager/pull/272#issuecomment-1047685536

Proposing this in draft mode for now.

1. Do we want the menu option back  (I know I do use it occasionally but will be happy to maintain my private version of this is nothing for mainline PM).
2. This introduces a DConf value to be saved per-user. While this is normal for sailfish apps to do, it; s a first for PM. Up until now all tunables were going form the UI -> Plugin -> Daemon -> `/etc/patchmanager2.conf`.  
This configuration value does not need to go through the daemon, as it doesn't need to know about it, it's UI-only. Also, that makes it multi-user ;)

### DIFF
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -22,6 +22,7 @@ Requires:   patch
 Requires:   grep
 Requires:   sed
 Requires:   sailfish-version >= 3.4.0
+Requires:   qml(Nemo.Configuration)
 BuildRequires:  pkgconfig(Qt5Core)
 BuildRequires:  pkgconfig(Qt5DBus)
 BuildRequires:  pkgconfig(Qt5Qml)
--- a/src/qml/PatchManagerPage.qml
+++ b/src/qml/PatchManagerPage.qml
@@ -1,7 +1,7 @@
 /*
  * Copyright (C) 2013 Lucien XU <sfietkonstantin@free.fr>
  * Copyright (C) 2016 Andrey Kozhevnikov <coderusinbox@gmail.com>
- * Copyright (c) 2021, Patchmanager for SailfishOS contributors:
+ * Copyright (c) 2021, 2022, Patchmanager for SailfishOS contributors:
  *                  - olf "Olf0" <https://github.com/Olf0>
  *                  - Peter G. "nephros" <sailfish@nephros.org>
  *                  - Vlad G. "b100dian" <https://github.com/b100dian>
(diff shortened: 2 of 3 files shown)


Olf0:
0. Oops, I obviously was not fully up to my usual quality standards, when I introduced what you fixed by [commit `cf87c44`](https://github.com/sailfishos-patches/patchmanager/pull/283/commits/cf87c44fc5a801d1ba978918178ff8b70dabb5ce).  Well, I know that the web-editor does add these superfluous spaces easily, but here I missed to pay attention.  Thanks for noticing and rectifying.
1. > Do we want the menu option back?
    
    As mentioned before, I am fine with having "Deactivate all Patches" permanently at the topmost position of the pulley menu of Patchmanager's main page, see [commit 272](https://github.com/sailfishos-patches/patchmanager/pull/272) for details (and and [issue 20](https://github.com/sailfishos-patches/patchmanager/issues/20) for the debate before).

2. > I know I do use it occasionally but will be happy to maintain my private version of this is nothing for mainline PM.

    This approach does not make much sense to me:
    * If you want (or need) it, likely somebody else also does.
    * I appears to be easier (in terms of workload) to implement a generic solution once, than to maintain a private branch in the long run.

3. > This introduces a DConf value to be saved per-user.  […]

    I expected this to be more intrusive, but it does not need that much code and does not look overly complicated.  As long as the default value is "off" I am fine with this.

    And the prospect of having a user specific configuration framework, which is only evaluated by the PM GUI application (in contrast to the PM daemon, which uses and must use a system-wide configuration in `/etc`), which also may be utilised for other settings of the PM GUI application later, is very nice.

Olf0:
Please check, that I denoted the technical details correctly when rephrasing the comment section for clarity:
```
    /*
     * The usual, system-wide configuration values are set via D-Bus plugin by the
     * Patchmanager daemon, which stores them in /etc/patchmanager2.conf
     * This configuration group "uisettings" is for settings which *solely* affect
     * the PM GUI application and consequently also are per-user settings.
    */
```
While the string changes (just spelled out, what was meant / already there) bear much less risk of mishaps, please check them, too: https://github.com/sailfishos-patches/patchmanager/pull/283/files#diff-c092975afb5d44c1002edfec699cb52c26bf8df373b6ed0696c0f8da182d889bR92-R93

Olf0 (approved):
AFAICT, this is reasonable and looking good.

Disclaimer: Please remember that "I am not a QML coder (IANAQC)!"

nephros:
> **A single "Check & Balance" question:** Is `Nemo.Configuration 1.0` available in SailfishOS 3.4.0 (i.e., the oldest SFOS release supported by recent Patchmanager versions)?

Definitely.  
It has been the standard way of doing settings in QML for ages. The corresponding package name is `nemo-qml-plugin-configuration-qt5`, and it is documented at https://sailfishos.org/develop/docs/nemo-qml-plugin-configuration/

PR #290, opened 2022-02-28
spec file: Add `Requires: %{name}` to subpackage

Olf0:
… `testcases`, because they do depend on Patchmanager to run.

### DIFF
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -46,11 +46,12 @@ Requires:   %{name}
 Requires:   libsailfishapp-launcher
 
 %description testcases
-The package patchmanager-testcases installs a small test application and a couple of
-Patches which are helpful for testing the internal functioning of Patchmanager, e.g.,
-when changes were made to it.
-Note that utilising these test cases requires a thorough understanding how Patchmanager
-internally works and hence is most likely only useful for developers.
+The package patchmanager-testcases installs a small test application and a
+couple of Patches which are helpful for testing the internal functioning of
+Patchmanager, e.g., when changes were made to it.
+Note that utilising these test cases requires a thorough understanding how
+Patchmanager internally works and hence is most likely only useful for
+developers.
 
 # This description section includes metadata for SailfishOS:Chum, see
 # https://github.com/sailfishos-chum/main/blob/main/Metadata.md
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -39,10 +39,11 @@ BuildRequires:  pkgconfig(rpm)
 BuildRequires:  pkgconfig(popt)
 
 %package testcases
-Summary: Provides test cases for Patchmanager
-Group: Development
-BuildArch: noarch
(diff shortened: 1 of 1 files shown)


nephros (approved):
Well in theory you could have it installed and "use" the "app" and the "patches" without patchmanager proper, which is why I left the dep off.

I don't thing the dependency is strictly necessary or useful, but have no real objection to the change, so approving.

It should not ever depend on a version of PM though.

Olf0:
> Well in theory you could have it installed and "use" the "app" and the "patches" without patchmanager proper, which is why I left the dep off.

You mean, by applying the patch files via `patch` utility?
And the fact that the app can always run without Patchmanager?
 
> I don't think the dependency is strictly necessary or useful, but have no real objection to the change, so approving.

I think (see above) I now understand the "not strictly necessary" part, but "not useful" surprises
me: I considered *»in theory you could have it installed and "use" [it] without patchmanager proper«* as mostly academic.
Practically the Patchmanager Testcases app and accompanied set of Patches exist for testing Patchmanager, right?
 
> It should not ever depend on a version of PM though.

Absolutely agreed, that does not make any sense: While Patchmanager and Patchmanager Testcases will carry same the version number the way both packages are generated, there (currently) is no version dependency at all.  Only the *»not ever«* puzzles me, because if a json-Patch contains metadata (to test), which requires a certain minimal version of Patchmanager to be correctly evaluated, this may change.  But the again, it would simply mean that a specific test is failing.

I really do not want to simply discard your objection, especially because the test cases are "your baby".
OTOH I do want to avoid the situations like the one we observed with Vlad: "Why does starting the app do nothing except for telling that no Patches were deployed?"  Without Patchmanager installed this becomes even harder to comprehend.

P.S.: An "easy way out" would be a [`Recommends:` weak dependency](https://weldr.io/RPM-Dependencies/#weak-dependencies-recommends-suggests-supplements-enhances), which effectively does nothing on SailfishOS.  That needs `rpm` ≥ 4.13 (to be recognised as a keyword) and SFOS 3.2.1 deploys v4.14.
But as it is effectively a no-op, I really want to be sure that this is the right thing to do.

nephros (approved):

nephros:
> > Well in theory you could have it installed and "use" the "app" and the "patches" without patchmanager proper, which is why I left the dep off.
> 
> You mean, by applying the patch files via `patch` utility? And the fact that the app can always run without Patchmanager?
> 

Precisely.

> > I don't think the dependency is strictly necessary or useful, but have no real objection to the change, so approving.
> 
> I think (see above) I now understand the _»not strictly necessary«_ part, but _»not useful«_ surprises me: I would consider the statement _»in theory you could have it installed and "use" [it] without patchmanager proper«_ as academic (as "in theory" indicates). Practically the Patchmanager Testcases app and accompanied set of Patches exist for testing Patchmanager, right?

Academic is a good word for my "objections", which is why I just hit the merge button.

(I had a longer reply for this typed up, but it got lost in a fatfingered reload and I'm not typing it again, sorry.)

PR #296, opened 2022-03-04
[web catalog  migration] define compile-time variables in .spec

nephros:
To prepare for a possible move of the Web Catalog infrastructure, make the core variables configurable/defined in the .spec file

### DIFF
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -14,9 +14,9 @@
 # CATALOG_URL SERVER_URL "/" API_PATH
 # MEDIA_URL   SERVER_URL "/" MEDIA_PATH
 
-%define catalog_server https://coderus.openrepos.net
-%define catalog_api_uri pm2/api
-%define catalog_defines SERVER_URL=%{catalog_server} API_PATH=%{catalog_api_uri}
+%global catalog_server https://coderus.openrepos.net
+%global catalog_api_uri pm2/api
+%define catalog_defines -DSERVER_URL=%{catalog_server} -DAPI_PATH=%{catalog_api_uri}
 
 Name:       patchmanager
 
@@ -127,7 +127,7 @@ Url:
 %build
 
 %qtc_qmake5 "PROJECT_PACKAGE_VERSION=%{version}"
-%qtc_make %{?_smp_mflags} %{catalog_defines}
+%qtc_make %{?_smp_mflags} EXTRA_CFLAGS="$CFLAGS %{catalog_defines}"
 
 
 %install
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -13,7 +13,7 @@
 # CATALOG_URL SERVER_URL "/" API_PATH
 %global catalog_server https://coderus.openrepos.net
 %global catalog_api_uri pm2/api
-%define catalog_defines -DSERVER_URL=%{catalog_server} -DAPI_PATH=%{catalog_api_uri}
+%global catalog_defines -DSERVER_URL=%{catalog_server} -DAPI_PATH=%{catalog_api_uri}
 
 Name:       patchmanager
 
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -5,15 +5,12 @@
 %{!?qtc_make:%define qtc_make make}
 %{!?qmake5_install:%define qmake5_install make install INSTALL_ROOT=%{buildroot}}
 
-# override defines:
-# defined in src/bin/patchmanager-daemon/patchmanagerobject.h
-#            src/qml/webcatalog.h:
-#
+# Override these defines in both (!):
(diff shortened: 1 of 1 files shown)


Olf0:
Well I see aspects which speak for putting all this into a single repository (API changes in sync, a single location for everything), but I assume this to be much larger and intrusive than the test cases.  Hence the Web Catalog may clutter this repository, because there would be three sub-projects in it, then.
Hence I think we should at least consider to put that into a new repository under `sailfishos-patches`, which you should be able to open at https://github.com/sailfishos-patches

As usual, these are just checks & balances questions.  If you name a couple of convincing reasons why "all in one repo" is better, my questioning was for properly documenting these reasons here, which is fine.  If not, we open a new repo and that is fine, too.

Olf0 (approved):
Looks fine for me.

Olf0:
@nephros, please revert my commit [219bc17](https://github.com/sailfishos-patches/patchmanager/pull/296/commits/219bc17c17e102f3b150ea4060f856113c73ff03) if it was wrong to eliminate the C header file define `# MEDIA_URL   SERVER_URL "/" MEDIA_PATH` copied as a comment here in the spec file.

Side note: Checking these  in `src/bin/patchmanager-daemon/patchmanagerobject.h` (and as a verbatim copy in `src/qml/webcatalog.h`) …
```
# SERVER_URL          "https://coderus.openrepos.net"
# API_PATH            "pm2/api"
# PROJECTS_PATH       "projects"
# PROJECT_PATH        "project"
# RATING_PATH         "rating"
# FILES_PATH          "files"
# MEDIA_PATH          "media"
# CATALOG_URL         SERVER_URL "/" API_PATH
# MEDIA_URL           SERVER_URL "/" MEDIA_PATH
```
… made me wonder, if there is a path element `pm2/` missing from `PROJECTS_PATH` and `PROJECT_PATH`, plus maybe also from `RATING_PATH`, `FILES_PATH` and `MEDIA_PATH`?
Because as a matter of fact `PROJECTS_URL="${SERVER_URL}/pm2/$PROJECTS_PATH"` and `PROJECT_URL="${SERVER_URL}/pm2/$PROJECT_PATH$PROJECT"`!  I have not checked the other three paths.

nephros:
> … made me wonder, if there is a path element `pm2/` missing from `PROJECTS_PATH` and `PROJECT_PATH`, plus maybe also from `RATING_PATH`, `FILES_PATH` and `MEDIA_PATH`?

No, as they are used like e.g. at https://github.com/sailfishos-patches/patchmanager/blob/master/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L1923:

     QUrl url(QStringLiteral(CATALOG_URL "/" PROJECTS_PATH));

so if anything they should have an `/api` too much. But I guess there's normal web browsing at `server_root/pm2/project/foo`, and JSON (or other) responses at `server_root/pm2/api/project/foo`:

 - https://coderus.openrepos.net/pm2/projects/
 - https://coderus.openrepos.net/pm2/api/projects

nephros:
> @nephros, please revert my commit [219bc17](https://github.com/sailfishos-patches/patchmanager/pull/296/commits/219bc17c17e102f3b150ea4060f856113c73ff03) if it was wrong to eliminate the C header file define `# MEDIA_URL SERVER_URL "/" MEDIA_PATH` copied as a comment here in the spec file. It looked unrelated and superfluous / potentially confusing to me here.

Well, the media path is actually a different beast than app the api locations.  
It's basically where the files are downloaded from (which the web catalog software allows to be somewhere different than the catalog itself).
   
It's fine to remove from the comments as you did, as we don't mention it otherwise in the .spec, but that define should probably become configurable in the spec as well.

Olf0 (commented):
I don't know why this MR threw a notification for me, as I cannot detect any status change etc.

Anyway, I reread our old conversation, and as always you answered all my checks&balances-questions exhaustively.  I suppose that is why I marked it as reviewed.

Now, with some newly gathered spec file know-how, I might ask the nitpicking question, why you decided not to specify all three new defines as `%global`:
* Generally defines via `%global` are preferred today, because …
* `%global` defines are evaluated once, when they are created.  In contrast, `%define`s are evaluated each time they are used; hence the are necessary when variables / other defines may change during while a spec file is worked through (by RPM, `libzypp` etc.), but only then.
* `%global` defines can be used in all sections of a spec file, but a `%define` in the header section is not valid in any of the scriptlets (`%pre`, `%post`, `%preun`, `%postun`, `%pretrans`, `%posttrans` and the `%filetrigger…`-scriplets).

Thus you may consider to change the only `%define` also to `%global`, if I did not miss an aspect.
Or not, because this is just a suggestion (from me) / general guidance (from the RPM maintainers / RedHat / Fedora, with a large overlap between these three).

HTH

Olf0 (approved):
IMO this is ready to merge.

Issue #277, opened 2022-02-22
Allow saving and loading activation state


nephros:
See also https://github.com/sailfishos-patches/patchmanager/pull/272#issuecomment-1047192225

For users who have a lot of patches it would be useful to have the option to restore a certain configuration of active patches.

So one can get out of failure mode quickly and with minimal hassle, and just in general.

For the first iteration I think saving a single configuration is enough, managing a list of favorite combinations is not necessary.

Olf0:
That would be really cool, I considered to document that multiple times in the last five years.  I did refrain from doing so, because it believe it is not easy to implement: The user may have uninstalled, updated or installed new Patches since writing that list.

I see two basic approaches:
1. Either that list is permanently kept up-to-date by Patchmanager (i.e., the list is updated after activating or deactivating any Patches, but also after uninstalling an activated or updating an activated Patch, assuming only activated Patches are denoted on this list) and handled transparently, i.e., there is simply a "Restore former activation state of all Patches" after issuing a `patchmanager --unapply-all` at the command line or Patchmanager detecting a failure and entering "failure mode" (which also deactivates all Patches), but no other user visible interface at the GUI.
2. Or the user explicitly saves and restores this list, but then there might be a different set of Patches and Patch versions installed when the restore action takes place, which must be handled well.

IMO approach 1 is much more usable, but might be harder to implement.

nephros:
The list of currently active patches is already saved in the config file and updated when anything about activation state changes.

See the `applied` key in the `[settings]` section of `/etc/patchmanager2.conf`.

What remains is to improve the "go out of failure state" actions to restore state correctly.

Olf0:
> The list of currently active patches is already saved in the config file and updated when anything about activation state changes.
> See the `applied` key in the `[settings]` section of `/etc/patchmanager2.conf`.

This is what I also realised after thinking about it: The "applied" state must already be stored somewhere.

> What remains is to improve the "go out of failure state" actions to restore state correctly, and the option to restore the state from that list on demand (i.e. after unapply all).

:+1:

Issue #204, opened 2021-11-14
Secret functionality: --reset-system


nephros:
The daemon supports a "secret" switch `--reset-system` which from cursory lookover will

 - unapply all patches
 - find all RPMs which own files which were affected by patches
 - refresh all repositories
 - reinstalls those packages

https://github.com/sailfishos-patches/patchmanager/blob/master/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L821

Qs:  

 - Is this a useful function we want to keep around?
 - Does it work as it is supposed to?

It's a nifty thing if it works, but as it does unattended package management I regard it as highly dangerous. And I'm not sure it's still required in the world of PM3 with the preload method.

nephros:
Additionally, now that we have 32/64 bit mangling, it might fail to find the correct paths, that would have to be looked into.

nephros:
This feature has been "deprecated" in [#527](https://github.com/sailfishos-patches/patchmanager/pull/527)

Issue #198, opened 2021-11-09
daemon: add option to list currently applied patches, and patches wanting service restart


nephros:
patchmanager-daemon offers the cmdline switches `-a`, `-u`, and `--unapply-all` to apply/unapply patches from scripts.

## Feature Request 1:

`-l|--list-applied`

Output the list of applied patches in a parseable format, so scripts can save (and possibly restore) that information.  
Useful e.g. in a system update situation, for when developing patches.

## Feature Request 2:

`-C|--check-services-restart`

Output whether PM needs a "Services Toggle". This is useful to know before applying/unapplying patches

nephros:
Note: both things are possible through DBus calls, but those are pain to script and parse.

Olf0:
> ## Feature Request 2:
> 
> `-C|--check-services-restart`

May I suggest to solely use lower case options, here `-c` instead of `-C`.

> Note: both things are possible through DBus calls, but those are pain to script and parse with plain unix tools.

What is so hard to parse, i.e., can you provide an example?
I may be able to support you a bit, but do not fully comprehend the issue yet (and generally per "grep, sed & awk" almost everything can be parsed; I think we should avoid "flex & bison", being mighty but hard to use).

nephros:
After having dabbled in this for a bit it turns out that doing this transforms the patchmanager binary from a dbus daemon with minimal interactive interface into a dbus client-and-server hybrid.

Maybe it would be a better design to implement controlling functions into a separate patchmanager-manager tool which is just a client.

nephros:
> What is so hard to parse, i.e., can you provide an example?
> I may be able to support you a bit, but do not fully comprehend the issue yet (and generally per "grep, sed & awk" almost everything can be parsed, although they are unsuitable for really parsing XML / HTML; I think we should avoid "flex & bison", being mighty but hard to use).

Thanks, but my comment there is more of a general complaint about the verbosity and amount of boilerplate that is necessary to interact with dbus.

It's all *doable* but rather unwieldy using classic Unix CLI.

Olf0:
Just a high-level perspective / feedback on the D-Bus topic:
* **You** determine the interface the PM-daemon uses.  While Andrey K. implemented this using D-Bus, I could imagine to switch to UNIX sockets for communicating with the daemon.  Though, one of the reasons why D-Bus exists is to overcome the drawbacks of classic (local) UNIX socket communication.
[...]

PR #302, opened 2022-03-18
multi-apply-tool: first version

nephros:
Inspired by #277 (and giving up expanding the dbus-calls in the binary for this), this little tools makes life easier for users handling large amounts of patches.

Features:

 - takes a list of patch names, and either activates or deactivates them 
 - rootless mode: can read and write list of patches to be de/activated as user
 - rootless mode: if run as non-root, output script file with corresponding actions
 - read list of patches to be de/activated from pm config file
 - read list of patches to be de/activated from user-specified file
 - save list of currently activated patches to user-specified file

Nothing magic, but useful.

### DIFF
--- a/src/tools/pm-multi-tool
+++ b/src/tools/pm-multi-tool
@@ -1,0 +1,71 @@
+#!/usr/bin/env sh
+
+PM_CONF=/etc/patchmanager2.conf
+
+function usage() {
+        printf "Patchmanager Multi-Patch operation tool.\n\n"
+        printf "USAGE: %s [-A|-U|-h] [-f FILE]\n\n" $0
+        printf "\t-A | --apply\t\t\tactivate/apply all patches marked active\n"
+        printf "\t-U | --unapply\t\t\tde-activate/unapply all patches marked active\n"
+        printf "\t-f | --file <filename>\t\tread list of patches from file\n"
+        printf "\t-e | --export <filename>\texport list of active patches to file\n"
+        printf "\t-h | --help\t\t\tprint this help\n"
+        printf "\nFILE must be a plain text file containing the (internal) names of patches, separated by white space\n"
+        printf "\n-A and -U options require you to be root to do something, and will simulate the actions otherwise\n"
+}
+function write_list() {
+        awk -F=  '/^applied/ {gsub(/\,/,"") ;  print $2} ' "${PM_CONF}" > "${1}"
+        exit 0
+}
+
+operation="-u"
+case $1 in
+    -A|--apply)
+        operation="--apply"
+    ;;
+    -U|--unapply)
+        operation="--unapply"
+    ;;
+    -e|--export)
+            if [ ! -w $(dirname "$2") ]; then
+            printf "Can not write to given location: %s. Exit.\n" "${2}" >&2
+            exit 2
+        fi
+        if [ -e "$2" ]; then
+            printf "File %s exists, will not overwrite. Exit.\n" "${2}" >&2
+            exit 2
+        else
+            write_list $2;
+            exit 0;
+        fi
+    ;;
+    -h|--help)
+        usage
+        exit 0
+    ;;
+    *) usage; exit 1 ;;
+esac
+
+plist=""
+
+case $2 in
+    -f|--file)
+        plist=$(< $3)
+    ;;
+    *)
+        plist=$(awk -F=  '/^applied/ {gsub(/\,/,"") ;  print $2} ' "${PM_CONF}")
+    ;;
+esac
+
+if [ ! $EUID -eq 0 ]; then
+        printf "Not running as root, printing command list only.\nYou can redirect this into a file and execute that as root.\n\n" >&2
+        printf "#!/usr/bin/env sh\n\n"
+fi
+for p in ${plist}; do
+  if [ $EUID -eq 0 ]; then
+    /usr/sbin/patchmanager $operation $p
+    sleep 0.2
+  else
+      printf "/usr/sbin/patchmanager %s %s; sleep 0.2\n" $operation $p
+  fi
+done


nephros:
Example use case:

  -  while testing some terrible developer's history app, a tester notices that this crashes lipstick.
  - this makes PM go into failure state, forcing them to manually re-activate all patches
  - which is majorly annoying for users whi have a lot of patches, and makes them want to stop testing the broken app
  - with multi-tool, the tester can now conveniently save the list of favourite patches into a backup file: `pm-multi-tool -e ~/patches.list`, and contiune the lipstick-crashing tests: 
  - some crashes later, they can restore their patch configuration: `pm-multi-tool -A -f ~/patches.list`

Olf0:
Questions:
- Where does `$EUID` come from?
- `write_list()` is only used once and a single line function → Better inline it, or is there another reason?
- I am not a fan of capital letter options.  I understand, that this is intended to indicate that they should be executed as root, right?   But for Patchmanager proper all options are lower-case, specifically the `-a` and `-u` options.  Hence I believe they should stay lower-case for consistency.  If you are O.K. with that, I will do that.
- Why is the variable `operation` preset to `-u` when the `case` statement covers all cases? → Preset it to "".
- More commonality of the executable name with `patchmanager` executable, e.g., `patchman-multi` or `patcher-multi`.
- I considered [my former write-up](https://github.com/sailfishos-patches/patchmanager/issues/198#issuecomment-964160584), but many aliased options are not nice.  Ideas?
- Do the long-opts [`--apply`](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-3e445eb245c9551bb1730fd84c4c1253da595cdd1eecdb380566d12a6bea4e62R24) and [`--unapply`](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-3e445eb245c9551bb1730fd84c4c1253da595cdd1eecdb380566d12a6bea4e62R27) really work for `patchmanager` proper?  Then please also [name them here](https://github.com/sailfishos-patches/patchmanager/issues/198).  We should also document all available options somewhere, including https://github.com/sailfishos-patches/patchmanager/issues/204

nephros:
> 7\. Do the long-opts [`--apply`](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-3e445eb245c9551bb1730fd84c4c1253da595cdd1eecdb380566d12a6bea4e62R24) and [`--unapply`](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-3e445eb245c9551bb1730fd84c4c1253da595cdd1eecdb380566d12a6bea4e62R27) really work for `patchmanager` proper?  Then please also [name them here](https://github.com/sailfishos-patches/patchmanager/issues/198).  We should also document all available options somewhere, including [Secret functionality: --reset-system #204](https://github.com/sailfishos-patches/patchmanager/issues/204)

Argh! yes! 
I have a patch somewhere which implements these, but haven't submitted it yet. Will clean it up and do another PR on them.

nephros:
> 1. Where does `$EUID` come from?

POSIX I think. Might be a GNU-ism.
It's "effective" UID, i.e. the UID a process is running as (as opposed to UID, the ID of the user running something). E.g. if user mark runs a setuid program whose executable file is owned by mary, UID is mark and EUID is mary.

nephros:
No objections to these three:

> * I am not a fan of capital letter options.  I understand, that this is intended to indicate that they should be executed as root, right?   But for Patchmanager proper all options are lower-case, specifically the `-a` and `-u` options.  Hence I believe they should stay lower-case for consistency.  If you are O.K. with that, I will do that.

My thought was as they are potentially "destructive", including one more keypress id a bit of a safeguard. Also, the original idea came from modifying the original patchmanager program, where there is `-a` and `-u` for one patch, and it makes sense there for the capitalized version to mean "all".

> * Why is the variable `operation` preset to `-u` when the `case` statement covers all cases? → Preset it to "".
> * More commonality of the executable name with `patchmanager` executable is desirable (to be found per tab-completion etc.), e.g., `patchman-multi` or `patcher-multi`.

Will do!

Olf0:
> > 7. Do the long-opts [`--apply`](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-3e445eb245c9551bb1730fd84c4c1253da595cdd1eecdb380566d12a6bea4e62R24) and [`--unapply`](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-3e445eb245c9551bb1730fd84c4c1253da595cdd1eecdb380566d12a6bea4e62R27) really work for `patchmanager` proper?  Then please also [name them here](https://github.com/sailfishos-patches/patchmanager/issues/198).  We should also document all available options somewhere, including [Secret functionality: --reset-system #204](https://github.com/sailfishos-patches/patchmanager/issues/204)
> 
> Argh! yes! I have a patch somewhere which implements these, but haven't submitted it yet. Will clean it up and do another PR on them.

But then (as currently only `-a`, `-u`, `--unapply-all´ and `--reset-system` are defined, we may really go for
1. `-a`, `--activate`
2. `-d`, `--deactivate` and for backward compatibility `-u`
3. `--deactivate-all` and for backward compatibility `--unapply-all`

That would make the terms consistent with the ones used at the GUI.

Olf0:
Idea to unify the options: Make `-e` just an "export" option, which writes to STDout, when no `-f FILE` is provided.  With `-f FILE` it writes to FILE.

Additional idea: `-a` and `-d`/`-u` read from STDin (oh, I realise that forgot how to do this in a shell script), when no `-f FILE` option is set.  It only stubbornly activate *all* Patches, when there is no input at STDin.

Olf0:
I hope you like my overhauling it.  Sorry, I became (side-) tracked into it.  I rather would have posed a PR, but started at the GitHub web-frontend, which does not allow for that.

I think I implemented all my suggestions.  I plan to review / revisit it in a couple of days.

Olf0:
Didn't we document somewhere what the allowed character set for (internal) Patch names is?
I did not find it, in order to check if my guess was O.K.: https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-2019b587d28f91dde2f21f3b67a4c4bace388705e450756d82a5b4b02cd6a6abR153

nephros:
> Didn't we document somewhere what the allowed character set for (internal) Patch names is? I did not find it, in order to check if my guess was O.K., [see line 153](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-2019b587d28f91dde2f21f3b67a4c4bace388705e450756d82a5b4b02cd6a6abR153).

Not that I remember, but I guess that's checked in the web catalog code in some way.

Olf0:
> I plan to review / revisit it in a couple of days.

Done.

Olf0 (approved):
Well, it is O.K. from my perspective, because I did the last changes.

Olf0 (changes requested):
Please re-review @nephros.

<sup>Simulating this by "request changes".</sup>

Olf0 (approved):

Olf0:
> > Didn't we document somewhere what the allowed character set for (internal) Patch names is? I did not find it, in order to check if my guess was O.K., [see line 153](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-2019b587d28f91dde2f21f3b67a4c4bace388705e450756d82a5b4b02cd6a6abR153).
> 
> Not that I remember, but I guess that's checked in the web catalog code in some way.

I was unable to find the code which does that at https://github.com/sailfishos-patches/pm-web-catalog.  Please [check that my checks are not too strict or too relaxed](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-2019b587d28f91dde2f21f3b67a4c4bace388705e450756d82a5b4b02cd6a6abR148); if anything than rather "too relaxed", because this also checks RPM Patches.  But either you or me (or both) should compare this to the Web Catalog, which is more or less a reference.

nephros:
> Please re-review @nephros.
> 
> Simulating this by "request changes".

Looks good, I can't review my own PR :)

Olf0:
@nephros, the last open point is the one below.  Please check.  If that results in a need for changes, you are welcome to delegate that to me.  Then just drop a pointer to the corresponding check in the Web Catalog's code here.

> > > Didn't we document somewhere what the allowed character set for (internal) Patch names is? I did not find it, in order to check if my guess was O.K., [see line 153](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-2019b587d28f91dde2f21f3b67a4c4bace388705e450756d82a5b4b02cd6a6abR153).
> > 
> > 
> > Not that I remember, but I guess that's checked in the web catalog code in some way.
> 
> I was unable to find the code which does that at https://github.com/sailfishos-patches/pm-web-catalog. Please [check that my checks are not too strict or too relaxed](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-2019b587d28f91dde2f21f3b67a4c4bace388705e450756d82a5b4b02cd6a6abR148); if anything then rather "too relaxed", because this also checks RPM Patches. But either you or me (or both) should compare this to the Web Catalog, which is more or less a reference.

nephros:
> @nephros, the last open point is the one below. Can you please check this? If that results in a need for a change, you are welcome to delegate that to me. Then just drop a pointer to the corresponding check in the Web Catalog's code here.
> 
> > > > Didn't we document somewhere what the allowed character set for (internal) Patch names is? I did not find it, in order to check if my guess was O.K., [see line 153](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-2019b587d28f91dde2f21f3b67a4c4bace388705e450756d82a5b4b02cd6a6abR153).
> > > 
> > > 
> > > Not that I remember, but I guess that's checked in the web catalog code in some way.
> > 
> > 
> > I was unable to find the code which does that at https://github.com/sailfishos-patches/pm-web-catalog. Please [check that my checks are not too strict or too relaxed](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-2019b587d28f91dde2f21f3b67a4c4bace388705e450756d82a5b4b02cd6a6abR148); if anything then rather "too relaxed", because this also checks RPM Patches. But either you or me (or both) should compare this to the Web Catalog, which is more or less a reference.

Thanks for the ping.

I looked a bit through the catalog code, all I can find are checks for archive file endings, and the "project name".

1. doing a `check_output()` when calling the unpack functions for an uploaded archive:

 - https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/filehandler.py#L16
 - https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/forms.py#L58

so I guess for uploaded archives anything goes that doesn't have something particularly weird as a file name.  
The unpack tests only test archives and do not attempt to write any files, so if the archive contains  "illegal" file names that would not be detected.  

2. The upload itself writes to `tmp/%s` so any illegal characters in the archive file would fail here, and the archive wouldn't upload successfully.

3. Now, *project names* are checked here: https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/models.py#L23 and allow `^([\w-]*)$'`. I think this is the relevant check, and yours is more specific than that, but it looks like they should be able to "co-exist" just fine.  
There may be some very subtle differences in what `\w` means in python vs. the matches in grep, but I doubt it will matter.

Olf0:
> > @nephros, the last open point is the one below. Can you please check this? If that results in a need for a change, you are welcome to delegate that to me. Then just drop a pointer to the corresponding check in the Web Catalog's code here.
> > > > > Didn't we document somewhere what the allowed character set for (internal) Patch names is? I did not find it, in order to check if my guess was O.K., [see line 153](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-2019b587d28f91dde2f21f3b67a4c4bace388705e450756d82a5b4b02cd6a6abR153).
> > > > 
> > > > Not that I remember, but I guess that's checked in the web catalog code in some way.
> > > 
> > > I was unable to find the code which does that at https://github.com/sailfishos-patches/pm-web-catalog. Please [check that my checks are not too strict or too relaxed](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-2019b587d28f91dde2f21f3b67a4c4bace388705e450756d82a5b4b02cd6a6abR148); if anything then rather "too relaxed", because this also checks RPM Patches. But either you or me (or both) should compare this to the Web Catalog, which is more or less a reference.
> 
> I looked a bit through the catalog code, all I can find are checks for archive files, and the "project name".
> 1. doing a `check_output()` when calling the unpack functions for an uploaded archive:
>     * https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/filehandler.py#L16
>     * https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/forms.py#L58
> 
>    so I guess for uploaded archives anything goes that doesn't have something particularly weird as a file name. The unpack tests only test archives and do not attempt to write any files, so if the archive contains "illegal" file names that would not be detected.
>  2. The upload itself writes to `tmp/%s` so any illegal characters in the archive file would fail here, and the archive wouldn't upload successfully.
>  3. Now, _project names_ are checked here: https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/models.py#L23 and allow `^([\w-]*)$'`. I think this is the relevant check, and yours is more specific than that, but it looks like they should be able to "co-exist" just fine.
>      There may be some very subtle differences in what `\w` means in python vs. the matches in grep (what with unicode and all that), but I doubt it will matter in practice.

I checked the last aspect first, i.e., if there are *"subtle differences in what `\w` means in python vs. the matches in grep (what with unicode and all that)"*, by reading [docs.python.org/3/howto/regex](https://docs.python.org/3/howto/regex.html) and [docs.python.org/3/library/re](https://docs.python.org/3/library/re.html).  AFAIU, the baseline is, "do not use `\w`, if one does not want a **locale** specific interpretation of the character class", i.e. here, write `^([\w-]*)$'` as `^([a-zA-Z0-9_-]*)$'`.

For *"yours is more specific than that, but it looks like they should be able to "co-exist" just fine"* in point 3: Well, mine is more specific (must start with `[a-zA-Z]` and end with `[a-zA-Z0-9]`), but also allows for `.+` in between (in addition to [a-zA-Z0-9_-]), plus mine must be at least two characters long.  IMO these are improvements, hence I will pose a PR for `pm-web-catalog/…/models.py`: https://github.com/CODeRUS/django-test/pull/8

For the general question, "What are we dealing with there?": To be honest I am a bit confused!  Are the [Patch names we are dealing with in our script](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-2019b587d28f91dde2f21f3b67a4c4bace388705e450756d82a5b4b02cd6a6abR148) what is called "*[`name' - Internal name](https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/forms.py#L13)" in the web-catalog?  I.e., are these the names of a downloaded Patchmanager-Patch as a whole (i.e., not unpacked)?
Isn't this exactly the file-name [filehandler.py](https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/filehandler.py#L7) handles?

I assume the answer to all three questions above is "Yes", but I am not sure.  Maybe @coderus can enlighten us with a brief statement on this.

CODeRUS:
iirc name is the final destination of unpack procedure. if you allow it to have some `/` or `.` it can possibly escape patches folder by doing `../..` etc and damage filesystem.

Olf0:
@CODeRUS, thank you very much for your prompt reply.

> iirc name is the final destination of unpack procedure. if you allow it to have some `/` or `.` it can possibly escape patches folder by doing `../..` etc and damage filesystem.

a. The RegEx I designed does not allow for using "`/`".  As "`/`" *and* "`.`" are necessary to construct "`../`", I cannot see any danger here.
    Still I appreciate much taking security consideration into account, because there may be someone trying to write a malicious Patch for Patchmanager.

b. As the *"name is the final destination of [the] unpack procedure"* (AFAICS [the "project name", which is validated](https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/models.py#L19-L47)), it is better to also validate the file-name of the uploaded file (["[models.py] Improve RegExes", ](https://github.com/CODeRUS/django-test/pull/8)https://github.com/CODeRUS/django-test/pull/8) and additionally the ones in the archive (["Validate file names", ](https://github.com/CODeRUS/django-test/pull/10)https://github.com/CODeRUS/django-test/pull/10).

CODeRUS:
can you somehow setup a test server to try all these before it will reach production? :)

Olf0:
Unfortunately not, I lack the infrastructure and know-how (never did anything with Django).

But the logic changes are minimal, just three more `if`-statements with the same fairly loose RegEx.  The biggest danger I see are typos breaking syntax (and one never detects one's own typos well).

I think thoroughly reviewing these two PRs will do, and rather provide a couple of days for that.  I will take another close look in tomorrow or on Tuesday (and denote when I did that), maybe @nephros can also take a took.   Or @nephros, do you have a staging instance of the web-catalog running?

CODeRUS:
~~you~~ someone :)

nephros:
> Or @nephros, do you have a staging instance of the web-catalog running?

Hm.

> I think I lost something in my basement, let me go upstairs and check!
> -- M.C.Escher

Yeah I had it running fir a while on a verry unstable and slowly-connected home server.

But I could provide a testing instance in a pinch.

Would an empty one (i.e. one with NO user accounts, no files uploaded etc) suffice, or would it be better to have a clone of the production catalog?

Olf0:
### 0. Background / Motivation
When I analysed in detail which checks for the file-names of uploads to Patchmanager3's web-catalog currently exist (i.e., the archive name and the names of the files in the archive), there was none (except for the one nephros already mentioned, [the archive file can be successfully written to the file-system](https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/filehandler.py#L28), plus [a file named `unified_diff.patch` exists in the uploaded archive](https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/filehandler.py#L28)).
While a checks for file name extensions are performed twice ([[1]](https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/models.py#L127) and [[2]](https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/filehandler.py#L16-L22), which makes sense, because [1] happens before the archive is written to the file-system and [2] [after that](https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/models.py#L133)) for the archive file (`tar.gz`, `.tar.bz2`, `.tar.xz` or `.zip`) and [for the files in the archive](https://github.com/sailfishos-patches/pm-web-catalog/blob/master/example/demo/filehandler.py#L29) ('.qml', '.js', '.png', '.svg', '.qm', plus the file-name 'unified_diff.patch') there are no checks for the full file names, specifically preventing a `../` to escape the working directory.  Hence I applied the simple RegEx I constructed for checking a Patch's "project name" (first [for the "multi-apply-tool"](https://github.com/sailfishos-patches/patchmanager/pull/302/files#diff-2019b587d28f91dde2f21f3b67a4c4bace388705e450756d82a5b4b02cd6a6abR148) originally discussed here, then also [for the "project name" supplied in the web-form for uploads to the web-catalog](https://github.com/CODeRUS/django-test/pull/8)) also to the [uploaded archive name and the file-names in the archive](https://github.com/CODeRUS/django-test/pull/10), which does not accept a slash ("/") in a name.

### 1. Reviewing the changes to Patchmanager3's web-catalog by "someone"
Task: Review of Python code changes and RegEx changes in the PRs ["[models.py] Improve RegExes", ](https://github.com/CODeRUS/django-test/pull/8)https://github.com/CODeRUS/django-test/pull/8 and ["Validate file names", ](https://github.com/CODeRUS/django-test/pull/10)https://github.com/CODeRUS/django-test/pull/10.

@pherjung, as you mentioned that you have some experience with Python code(ing), and these changes are really small (by amount of changed / added code), may I ask you to review these two PRs.  As already discussed and because I try to pull you into reviewing something you have not volunteered for (Patchmanager3's web-catalog), you sure have the freedom to say "No"; unfortunately I have no idea who else I may ask (CODeRUS lacks any time, nephros usually does not code in Python, I have written these changes).  To understand and check the motivation and logic behind these changes, start reading [at this comment of this thread here](https://github.com/sailfishos-patches/patchmanager/pull/302#issuecomment-1200289404) up to this message you are currently reading, and the commit comments for the individual commits in the two PRs.  Take your time for this review (i.e., this review is not temporally critical), and I would appreciate much, if you help to alleviate CODeRUS' (and to a lesser extent also my) concerns of deploying unreviewed changes in a production environment; no matter how small these changes are, typos (syntax or structural errors) or think'os (logic or reasoning flaws) may have slipped in.  Feedback is best provided as comments to these two PRs, maybe except for fundamental think'os, which are better discussed here so nephros can participate.

### 2. Testing the changes to Patchmanager3's web-catalog by "someone"
> > Or @nephros, do you have a staging instance of the web-catalog running?
> 
> Yeah I had it running for a while on a very unstable and slowly-connected home server. They are offline meanwhile though.
> But I could provide a testing instance in a pinch.
> Would an empty one (i.e. one with NO user accounts, no files uploaded etc) suffice, or would it be better to have a clone of the production catalog?

Because I solely altered code, which is executed when uploading a Patch to the web-catalog, I think that uploading a few extant Patches with the PRs ["[models.py] Improve RegExes", ](https://github.com/CODeRUS/django-test/pull/8)https://github.com/CODeRUS/django-test/pull/8 and ["Validate file names", ](https://github.com/CODeRUS/django-test/pull/10)https://github.com/CODeRUS/django-test/pull/10 employed is fully sufficient.  If you have an idea how to easily automate the uploading of Patches to the web-catalog, then you might clone the production catalog and re-upload all of them to the test instance.  As I mentioned to Patrick (pherjung), this is not temporally critical.

pherjung:
I'll take a look next week.

pherjung:
Speaking of a test instance, I have a VM running where I can do such job. I can add your SSH public key if you want to access that VM.

Olf0:
> Speaking of a test instance, I have a VM running where I can do such job. I can add your SSH public key if you want to access that VM.

Thank you for the offer, but I deliberately wanted to avoid learning how to install and configure Django and the Web Catalog software, because that would likely cost me some hours, which translates to at least a week in wall time.  I also remembered that nephros once had a test instance of the Web Catalog running, but did not know its status and accessibility: As nephros pointed out, [he was willing to reactivate it](https://github.com/sailfishos-patches/patchmanager/pull/302#issuecomment-1200503016), and the security, bandwidth and maybe also stability limitations are not relevant for testing these additional input filters.  Meanwhile I have gained access to it and yesterday we managed to deploy the current master branch plus my two PRs on it.  I will now consider how to test effectively and create another PR which overhauls the strings.

Still I would appreciate very much, if you thoroughly scrutinise my two PRs by reviewing the changes and their reasoning.  While I am quite convinced that these small changes are fine (minus own typos and think'os I might fail to detect), for a good part this is about assuring coderus that it is O.K. to deploy these changes in his production environment (because he lacks the time to test *and* to quickly react, if there is some fall-out).

Olf0:
@nephros, (well, actually I) ultimately intermingled two topics here: The *multi-apply-tool* and enhancements of the Web-Catalog.  While I hope to return to the latter in the second week of January 2023, IMO this MR can and should be merged now, because it has been ready for months.  I also [prepared a release of v3.2.3](https://github.com/sailfishos-patches/patchmanager/releases/tag/untagged-baae02fd397f98046009), because we already have [accumulated quite some commits](https://github.com/sailfishos-patches/patchmanager/compare/3.2.2...master), which luckily all look non-critical so this should be an easy release.

I was about to hit the "Squash and merge" button in GitHub's web-frontend, but first wanted align with you(r opinion).

nephros:
> IMO this MR can and should be merged now,
> [...]
> I was about to hit the "Squash and merge" button in GitHub's web-frontend for this MR, but first wanted align with you(r opinion); also about releasing v3.2.3 when this is merged.

Agreed, and I have nothing queued for this branch IIRC.

Therefore: LGTM, and merging.

Issue #301, opened 2022-03-17
Can not be installed on 4.4.0.58EA


nephros:
**SailfishOS VERSION** (Settings → About product → Build): 4.4.0.58 (EA)
<br />**HARDWARE** (Settings → About product → Manufacturer & Product name): unknown
<br />**PATCHMANAGER VERSION** (Settings → Patchmanager → [Top pulley] About): unknown

#### BUG DESCRIPTION

Leszek reported in the Telegram channel that:

1. the 4.4 update process removes patchmanager (which AFAIK is a first)
2. it can not be installed after the update because of missing dependencies

#### ADDITIONAL INFORMATION

Which dependencies are failing to resolve is currently unknown.

nephros:
The current version (3.2.2) has the following dependencies on SFOS 4.3 according to OBS:

```
config(patchmanager) = 3.2....	patchmanager
grep	gnu-grep busybox-symlinks-grep
ld-linux-armhf.so.3	glibc
ld-linux-armhf.so.3(GLIBC_2.4)	glibc
libQt5Core.so.5	qt5-qtcore
libQt5Core.so.5(Qt_5)	qt5-qtcore
libQt5DBus.so.5	qt5-qtdbus
libQt5DBus.so.5(Qt_5)	qt5-qtdbus
libQt5Gui.so.5	qt5-qtgui
libQt5Gui.so.5(Qt_5)	qt5-qtgui
libQt5Network.so.5	qt5-qtnetwork
libQt5Network.so.5(Qt_5)	qt5-qtnetwork
libQt5Qml.so.5	qt5-qtdeclarative
libQt5Qml.so.5(Qt_5)	qt5-qtdeclarative
libQt5Quick.so.5	qt5-qtdeclarative-qtquick
libQt5Quick.so.5(Qt_5)	qt5-qtdeclarative-qtquick
libc.so.6	glibc
libc.so.6(GLIBC_2.28)	glibc
libc.so.6(GLIBC_2.4)	glibc
libc.so.6(GLIBC_2.9)	glibc
libdl.so.2	glibc
libdl.so.2(GLIBC_2.4)	glibc
libgcc_s.so.1	libgcc
libgcc_s.so.1(GCC_3.5)	libgcc
libnemonotifications-qt5.so.1	nemo-qml-plugin-notifications-qt5
libpthread.so.0	glibc
libpthread.so.0(GLIBC_2.4)	glibc
librpm.so.9	rpm
[...]

PR #303, opened 2022-03-19
set CI build image to latest version

nephros:


b100dian (approved):
Fully agree to use the latest - but it leaves me wonder how do we do next releases in openrepos. Do we download OBS builds from 3.4/4.0 for 'legacy'? Or just recommend chum for <4.2? (that is, assuming that the 4.2 version added librpm.9).
Also, do you know why do we depend on librpm?:) @Olf0 maybe you can share your thoughts here

nephros:
> Also, do you know why do we depend on librpm

Introduced in 301e3108, it is used to get metadata about patches which have been installed by RPM. It's also used elsewhere by the daemon.

See https://github.com/sailfishos-patches/patchmanager/blob/master/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L1722

nephros:
> how do we do next releases in openrepos.

Oh I was under the impression Openrepos packages were already taken from OBS builds, and CI is just that, CI?

Olf0:
First of all: Establish some cross-linking!
**This is about trying to resolve issue #301 (if it really is an issue!).**

I already came to the conclusion that people may have not used the Patchmanager build for SFOS 4.3.0 from SailfishOS:Chum, because I do not believe that Jolla deliberately performs such a disruptive change from one SFOS release to the next one (accidentally the already did that).  Unfortunately the issue report lacks almost all relevant information, which is not @nephros' fault, but the fault of the undisclosed reporter at Telegram; I would have put this issue on a halt with the "info needed" tag.

> > > **set CI build image to latest version**

I do not like that (literally setting the build target to "**latest**") and would have "nack'ed" it, because that usually implies a "moving", hence unknown target.  Obviously @nephros came to the same conclusion.

But the now suggested build target `4.3.0.15` does not exist, thus I have to "nack" it, too!
See https://hub.docker.com/r/coderus/sailfishos-platform-sdk/tags

> > how do we do next releases in openrepos.
> 
> Oh I was under the impression Openrepos packages were already taken from OBS builds, and CI is just that, CI?

No, I did use the CI builds.  The same for the releases page here at GitHub.  Simply because they did work on all SFOS releases up to now.

That is why I was happy with SFOS 4.0.1 as a target: The docker image downloaded for each CI run is much smaller, it is the first SFOS release which supports `aarch64` (i.e., all three architectures currently supported by SFOS) and the resulting code supposedly still runs on SFOS 3.4.0, which is the oldest SFOS release Patchmanager ≥ 3.1 supports.

> Anyway it would be easy to add another Github workflow for release builds if that is necessary.

Generally "yes", but the real question is how to handle that at Openrepos.  I currently cannot think of any way out, besides either stating there, "Use SailfishOS:Chum for SFOS ≥ 4.4.0.", or "Use SailfishOS:Chum for SFOS < 4.{3|4}.0."

The change of `librpm.so.8`→`librpm.so.9` must have occurred in SFOS 4.3.0 earliest (or really 4.4.0; [lets see what @a-dekker replies](https://github.com/sailfishos-patches/patchmanager/issues/301#issuecomment-1073149773)), because the commits were too late for 4.2.0, see https://github.com/sailfishos/rpm/commits/master and https://coderus.openrepos.net/whitesoft/sailversion

Hence this MR cannot work in principle (as the 4.2.0 build target would not work and a 4.3.0 build target does not exist) as long as @CODeRUS or someone else does not create [a docker-based build target for 4.3.0.15](https://hub.docker.com/r/coderus/sailfishos-platform-sdk/tags).
Side note: @CODeRUS please do not understand this statement as a try to put pressure on you to create a docker image for 4.3.0.15 (I considered filing an issue report for [github-sfos-build](https://github.com/CODeRUS/github-sfos-build), but then refrained from doing so), even though numerous SFOS-software at GitHub would benefit from that.  Only do so if time & motivation permits, or (that might be better in the long run) create a guide "How to create a "SailfishOS docker-image for CI/CD-runs at GitHub" and leave performing the concrete task (this one *and* future ones) to others.

Olf0 (changes requested):
TL;DR: **NAK**

nephros:
I mixed up tooling target (which does have 4.3.0.15) with docker tag which does not. It's I think possible to have the CI use "latest" docker image and still specify the build taget to something different, but there have been enough reasons given not to.

Therefore, closing. "Personal Developer CI" builds can easily be done through GH fork, local docker run, or OBS with any version.

Olf0:
Sorry to ask, but I would like to understand your line of thoughts.

> I mixed up tooling target (which does have 4.3.0.15) with docker tag which does not.

Does "tooling target" either mean OBS or the `mb2`call in the [`build.yml` CI runner config](https://github.com/sailfishos-patches/patchmanager/blob/master/.github/workflows/build.yml), or something else?

> It's I think possible to have the CI use "latest" docker image 

… which is equivalent to 4.2.0.x …

> and still specify the build target to something different, 

Yes, but not any *higher* version, right?

I know it is a little bit confusing, but AFAIU Coderus back-translated Jolla's SDK version numbers to SFOS release version numbers, which is much more convenient and less error prone (not to accidentally pick the wrong one).  Still I would not expect any SDK release to compile software for SFOS releases which did not exist when the SDK release was made.

> but there have been enough reasons given not to.
> 
> Therefore, closing. "Personal Developer CI" builds can easily be done through GH fork, local docker run, or OBS with any version.

Well, if you see a way to build for SFOS 4.3.0 or even 4.4.0 with the extant docker images, I think we should pursue that.  And that ought to be tested before being deployed "in production".  Additional we may consider to keep the old CI runs for build testing and only do the new ones for releases ("on tag push").

Olf0:
BTW, I wonder why the RPM size of the CI builds at GitHub are roughly four times the size (1,8 - 2 MB) of the OBS builds (~ 480 KB)!?!

nephros:
> BTW, I wonder why the RPM size of the CI builds at GitHub are roughly four times the size (1,8 - 2 MB) of the OBS builds (~ 480 KB)!?!

Without looking at it in detail, most likely because both build with debug symbols enabled, but OBS strips them out in debug/debugsource packages while the build call at the CI does not.

If that is so, that could be changed, I think, either by doing

    mb2 -t SailfishOS-$RELEASE-aarch64 build --enable-debug;

or the rpmbuild environment has to be configured differently to do the debug/debugsource packages.

Olf0:
@nephros, any comments WRT https://github.com/sailfishos-patches/patchmanager/pull/303#issuecomment-1081000706?

Olf0:
> Hence this MR cannot work in principle (as the 4.2.0 build target would not work and a 4.3.0 build target does not exist) as long @CODeRUS or someone else does not create [a docker-based build target for 4.3.0.15](https://hub.docker.com/r/coderus/sailfishos-platform-sdk/tags). Side note: @CODeRUS please do not understand this statement as a try to put pressure on you to create a docker image for 4.3.0.15 (I considered filing an issue report for [github-sfos-build](https://github.com/CODeRUS/github-sfos-build), but then refrained from doing so), even though numerous SFOS-software at GitHub would benefit from that. Only do so if time & motivation permits, or (that might be better in the long run) create a guide »How to create a "SailfishOS docker-image for CI/CD-runs at GitHub« and leave performing the concrete task (this one for 4.3.0.15 _and_ future ones for ≥ 4.4.0.z) to others.

Thanks to @CODeRUS we now could use 4.4.0.58 as the CI target release.

Still I believe we should keep the CI as a simple and "slender" (4.0.1.45 is ~ 2,5 GB to download, but 4.4.0.58 is ~ 3,5 GB) build check, and keep utilising the a personal SailfishOS-OBS repository (=> `Vendor: meego`) for the builds we distribute at OpenRepos; SailfishOS:Chum builds there any way, but with the `Vendor:` field set to `chum`.  So this discussion here was fruitful to gain understanding of the pitfalls and create a well working distribution scheme.

PR #305, opened 2022-03-30
spec file: Pull comments out of %description sections

Olf0:
… because they are translated into empty lines, which shows when viewing packages from SailfishOS:Chum with the `rpm` utility.


Olf0:
… what reminds me of Clockwork Orange.  A cool and disturbing movie, which I really like.

PR #306, opened 2022-04-01
Add `--enable-debug` to reduce RPM size

Olf0:


Olf0:
As suggested in [this comment for PR 303](https://github.com/sailfishos-patches/patchmanager/pull/303#issuecomment-1085832841) and tested, this is working fine: Executables are now ~500 KB, plus ~2,5 MB debuginfo and ~60 KB debugsource.

PR #326, opened 2023-01-05
Use modern statements in spec file

Olf0:
- `%global` instead of `%define` for static expressions: `%global`s are evaluated once when set, `%define`s are evaluated each time used.  Plus the scope of `%global`s is all sections, including the scriptlets.
- `-n %{name}-%{version}` always has been the default for `%setup`, hence omitted now.
-  `rm -rf %{buildroot}` as first statement in the `%install` section is long obsolete, thus superfluous.

### DIFF
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -1,4 +1,4 @@
-%define theme sailfish-default
+%global theme sailfish-default
 
 # These macros should already be defined in the RPMbuild environment, see: rpm --showrc
 %{!?qtc_qmake5:%define qtc_qmake5 %qmake5}
@@ -37,6 +37,7 @@ BuildRequires:  pkgconfig(libshadowutils)
 BuildRequires:  qt5-qttools-linguist
 BuildRequires:  pkgconfig(rpm)
 BuildRequires:  pkgconfig(popt)
+
 
 %package testcases
 Summary:    Provides test cases for Patchmanager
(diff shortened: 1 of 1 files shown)


nephros:
I believe the `%setup -n foo` is required for OBS to work but have to check.

nephros:
When you say the rm on the bildroot is obsolete, does this mean it is done implicitly?

I know the `%clean` macro is obsolete, but cleaning the buildroot is useful when building locally.

nephros:
> believe the %setup -n foo is required for OBS to work but have to check.

Scratch that, works fine without it.

Olf0:
> I believe the `%setup -n foo` is required for OBS to work but have to check.

Definitely by OBS in general, but maybe by Jolla's `git_tar`: I once analysed it somewhere, it is a shell script. Will try to find my analysis.



> When you say the rm on the bildroot is obsolete, does this mean it is done implicitly?
> 
> I know the `%clean` macro is obsolete, but cleaning the buildroot is useful when building locally.

Yes, both the `%clean` macro and `rm -rf %{buildroot}` is long obsolete, according to the RPM documentation etc. (Fedora packaging guidelines an many more) and I checked that in many `rpmbuild` runs.

And it is successfully build by our CI workflow, which utilises Coderus' Sailfish-SDK images, hence `mb2` and all the other pieces of Jolla's SDK.

Olf0:
> all good.

I am still glad (every time), that you double-checked and triggered me to re-evaluate and properly document my reasoning.

PR #328, opened 2023-01-05
Fix Issue 325: compiler warnings

nephros:
Avoid `-Wunused-result`, `-Wunused-function`, and `-Wunused-parameter` warnings.

@b100dian should be non-critical, but as as the preload library has the potential to really cause a bad day for users, please review carefully.


b100dian (approved):
Looks good, thanks for taking care of this!

Olf0 (approved):
LGTM  (It is C, not C++, hence my opinion is more profound.)

PR #329, opened 2023-01-05
Treat warnings as errors

b100dian:
Since warnings seem to be gone, let's keep it this way.

(I converted a file with dos2unix, so you might need to append `?w=1` to the diff url)


Olf0 (approved):
LGTM

Issue #333, opened 2023-01-06
Split "Allow incompatible Patches" from "Developer Mode"


nephros:
"Developer Mode" currently disables (or rather relaxes) the version check, but it also does more, such as enabling the message about old patch.json formats in the Patch Details page.

We should split "Version Check" from "Developer Mode", and offer either as a config option.

_Originally posted by @nephros in https://github.com/sailfishos-patches/patchmanager/issues/322#issuecomment-1295056631_

nephros:
Well, that was easier than expected.

One thing remains: If users upgrade to this, their "developerMode" setting from an old version will map to "developerMode" in the new version, and "strictCompatability" will default to false.

While this is probably unexpected, it's "fail-safe", because the dangerous option falls back to a safe default.

We could rename "developerMode" to force both to defaults, and/or handle a migration for the now split semantics. Or not - is it important?

Olf0:
> "Developer Mode" currently disables (or rather relaxes) the version check, but it also does more, such as enabling the message about old patch.json formats in the Patch Details page.
> 
> We should split "Version Check" from "Developer Mode", and offer either as a config option.

I absolutely concur that the current "Developer Mode" shall be split into two separate user settings, as [I already denoted in issue 322](https://github.com/sailfishos-patches/patchmanager/issues/322#issuecomment-1305011167).

> Well, that was easier than expected.

Indeed, as your three commits show: [d2d8387](https://github.com/nephros/patchmanager/commit/d2d8387f0a38bff14d5e9ccf891e2ecf3c7387c6), [e712296](https://github.com/nephros/patchmanager/commit/e712296a6e4409e9eb2aba3866afb29645b4476c) and [9fa793e](https://github.com/nephros/patchmanager/commit/9fa793e25e9234e7cdb6ed93937648c3e108c08a).

I also like the fact, that this splits the issues users have been experiencing for so long with the original "Developer mode" from the developer facing issue #322.

But …
> One thing remains: If users upgrade to this, their "developerMode" (which we present as "Allow incompatible patches" in the UI) setting from an old version will map to "developerMode" in the new version, and "strictCompatability" will default to true (i.e. "Allow incompatible patches" is OFF).

[...]

PR #334, opened 2023-01-06
Split "Allow incompatible Patches" from "Developer Mode"

nephros:
Fixes: #333

### DIFF
--- a/src/qml/PatchManagerPage.qml
+++ b/src/qml/PatchManagerPage.qml
@@ -376,7 +376,7 @@ Page {
             function doPatch() {
                 view.model.saveLayout()
                 if (!patchObject.details.patched) {
-                    if (!PatchManager.strictCompatability || patchObject.details.isCompatible) {
+                    if ((PatchManager.sfosVersionCheck !== VersionCheck.StrictChecking) || patchObject.details.isCompatible) {
                         patchObject.apply()
                     } else {
                         errorMessageComponent.createObject(background, {text: qsTranslate("", "This Patch is incompatible with the installed SailfishOS version.")})
--- a/src/qml/PatchManagerPage.qml
+++ b/src/qml/PatchManagerPage.qml
@@ -376,7 +376,7 @@ Page {
             function doPatch() {
                 view.model.saveLayout()
                 if (!patchObject.details.patched) {
-                    if ((PatchManager.sfosVersionCheck !== VersionCheck.StrictChecking) || patchObject.details.isCompatible) {
+                    if ((PatchManager.sfosVersionCheck !== VersionCheck.Strict) || patchObject.details.isCompatible) {
                         patchObject.apply()
                     } else {
                         errorMessageComponent.createObject(background, {text: qsTranslate("", "This Patch is incompatible with the installed SailfishOS version.")})
--- a/src/qml/PatchManagerPage.qml
+++ b/src/qml/PatchManagerPage.qml
@@ -376,7 +376,7 @@ Page {
             function doPatch() {
                 view.model.saveLayout()
                 if (!patchObject.details.patched) {
-                    if (PatchManager.developerMode || patchObject.details.isCompatible) {
+                    if (!PatchManager.strictCompatability || patchObject.details.isCompatible) {
                         patchObject.apply()
                     } else {
                         errorMessageComponent.createObject(background, {text: qsTranslate("", "This Patch is incompatible with the installed SailfishOS version.")})
--- a/src/qml/SettingsPage.qml
+++ b/src/qml/SettingsPage.qml
@@ -60,7 +60,7 @@ Page {
         if (PatchManager.developerMode === true) {
             console.info("Migrating settings from deprecated developerMode setting.")
             PatchManager.patchDevelMode = true
-            PatchManager.sfosVersionCheck = VersionCheck.RecklessChecking
+            PatchManager.sfosVersionCheck = VersionCheck.NoCheck
             PatchManager.developerMode = false
         }
     }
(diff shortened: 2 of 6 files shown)


nephros:
Marking as Draft, pending changes requested in #333

Olf0 (commented):
I still have my reservations WRT this "easy implementation", [as already denoted](https://github.com/sailfishos-patches/patchmanager/issues/333#issuecomment-1373987714).

I feel I must at least insist on creating a seamless migration from the old `developerMode` setting to the new `strictCompatability` setting (value needs to be negated, something I tried to avoid in my suggestions) in order to avoid negative feedback from many Patchmanager users, when this change is released.

nephros:
@Olf0 sorry, I have the bad habit of using `fixup` and force-push in my "private" forks.
I overwrote one of your commits doing this.

Please hold off any pushes until I take this PR out of Draft mode.

Olf0:
> @Olf0 sorry, I have the bad habit of using `fixup` and force-push in my "private" forks. I overwrote one of your commits doing this.

That is fine: I have the bad habit of committing trivial change right away.

I created another private branch and a MR, but …

> Please hold off any pushes until I take this PR out of Draft mode.

… failed to update it twice: Will wait and continue to analyse docker@GitHub to achieve caching of the many GBs of docker-image data which are downloaded for each CI run.

nephros:
Okay, have not tested it yet, but it *looks* okay. Have at it!

nephros:
migration snippet needs work.

Dropdown is not set to "allow any".

Olf0:
I first hesitated a bit with [the migration logic](https://github.com/sailfishos-patches/patchmanager/pull/334/files#diff-c092975afb5d44c1002edfec699cb52c26bf8df373b6ed0696c0f8da182d889bR58-R66), but I cannot see any better way (and see that my original ideas were abstractly nice, but not implementable with the extant boolean setting `developerMode`; i.e., purely academic).

Hence I have only four minor commits to suggest via [MR \#3](https://github.com/nephros/patchmanager/pull/3).

Olf0 (approved):
IMO this is basically looking fine.

Still I would love to comprehend these two statements, either by being addressed by some commit or a brief explanation.  But never mind, I am aware these were just two thoughts being denoted late at night.

> migration snippet needs work.

As [stated above](https://github.com/sailfishos-patches/patchmanager/pull/334#issuecomment-1374356211), IMO its basic logic is O.K. and hard to design differently.

> Dropdown is not set to "allow any"

I simply fail to parse this statement, though [had an idea what it may address](https://github.com/sailfishos-patches/patchmanager/pull/334#issuecomment-1374356211).

nephros:
Pls do not merge yet.

While the migration logic is correct, the Combonox does not yet work correctly.

 1. Its menu has only 2 Entries, while the enum has 4. Therefore when selecting 'No Check' a value of 1 is saved where it should be 3.
 2. As the migration logic saves a 3, when loading the combobox it has no value for it, and falls back to -1/undefined, which is equivalent to the first entry in the menu.

I have workarounds for both but must test them first.

Olf0:
> Pls do not merge yet.

I sure won't, it is "your baby" (unless you have ceased to care about).

> While the migration logic is correct, the Combonox does not yet work correctly.
> 
>     1. Its menu has only 2 Entries, while the enum has 4. Therefore when selecting 'No Check' a value of 1 is saved where it should be 3.
> 
>     2. As the migration logic saves a 3, when loading the combobox it has no value for it, and falls back to -1/undefined, which is equivalent to the first entry in the menu.

Thanks for the details.

> I have workarounds for both but must test them first.

I will happily review them, when they are ready.

nephros:
Please see 4a96399.

I decided to do the simple solution, have no gaps in the value list.

I know this results in a slightly weird order of the menu options (Strict - None - Relaxed), if I had to argue this I'd say, 0 is always the default, and the others are sorted in order of decreasing strictness.

Anyway, doing it this way means we need no workarounds, and also no workarounds/migrations needed when new options are added.

I also have removed the planned "Careless" option for now - partly because you pointed out that in current SFOSes, two options would be superfluous, and partly because fine-grainedness of a "relaxed" option can be tuned somewhere else (i.e. an extra tunable) if necessary.

Olf0 (approved):
This looks fine to me.  So I guess you are all set, now.

nephros:
And like a true evil American Senator, I sneak in an almost unrelated UI fix in 1dc424b as well.

Olf0:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&Tab;&Tab;▲
Oh, good catch.&Tab;&Tab;│

nephros:
@olf you asked to not squash your commits in https://github.com/nephros/patchmanager/pull/2 and https://github.com/nephros/patchmanager/pull/3, do you want to keep them when merging this as well?

If yes, I'd clean up and rebase this branch first. If not I´d just squash-merge this as usual.

Olf0:
> @olf you asked to not squash your commits in [nephros#2](https://github.com/nephros/patchmanager/pull/2) and [nephros#3](https://github.com/nephros/patchmanager/pull/3), do you want to keep them when merging this as well?

Well, that is not really worth the effort, hence …

> If yes, I'd clean up and rebase this branch first. If not I´d just squash-merge this as usual.

… "just squash-merge this as usual".

PR #337, opened 2023-01-09
Really stick to new nomenclature, plus …

Olf0:
1. Fix help output in [`src/bin/patchmanager-daemon/main.cpp`, lines 51-61](https://github.com/sailfishos-patches/patchmanager/blob/26eb9bdb175241e6d85562d28b556a852f03aff9/src/bin/patchmanager-daemon/main.cpp#L51-L61).
2. Fix and slightly extend argument evaluation logic in [`src/bin/patchmanager-daemon/patchmanagerobject.cpp`, lines 941-953](https://github.com/sailfishos-patches/patchmanager/blob/26eb9bdb175241e6d85562d28b556a852f03aff9/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L941-L953).
3. For changes due to the new nomenclature, see, e.g., [Patchmanager's Wiki page *Terms-and-wording*, entry *"enable"*](https://github.com/sailfishos-patches/patchmanager/wiki/Terms-and-wording#enable).

### DIFF
--- a/.github/workflows/ci_sfos3.4.0-armv7hl.yml
+++ b/.github/workflows/ci_sfos3.4.0-armv7hl.yml
@@ -16,7 +16,7 @@ jobs:
   build:
     runs-on: ubuntu-22.04
     env:
-      SEGMENT_DOWNLOAD_TIMEOUT_MINS: 3
+      SEGMENT_DOWNLOAD_TIMEOUT_MINS: 1
     steps:
     - name: Checkout
       uses: actions/checkout@v3
--- a/src/bin/patchmanager-daemon/main.cpp
+++ b/src/bin/patchmanager-daemon/main.cpp
@@ -53,11 +53,11 @@ void help()
     std::cout << "patchmanager" << std::endl;
     std::cout << std::endl;
     std::cout << "Usage:" << std::endl;
-    std::cout << "  patchmanager                : Run as daemon" << std::endl;
+    std::cout << "  patchmanager                : Run Patchmanager as daemon" << std::endl;
     std::cout << "  patchmanager -a <Patch>     : Enable and activate a Patch" << std::endl;
     std::cout << "  patchmanager -u <Patch>     : Deactivate and disable (unapply) a Patch" << std::endl;
     std::cout << "  patchmanager --unapply-all  : Deactivate and disable (unapply) all Patches" << std::endl;
-    std::cout << "  patchmanager --daemon       : Daemonize" << std::endl;
+    std::cout << "  patchmanager --daemon       : Daemonize Patchmanager" << std::endl;
 }
 
 int main(int argc, char **argv)
--- a/src/bin/patchmanager-daemon/main.cpp
+++ b/src/bin/patchmanager-daemon/main.cpp
@@ -53,11 +53,11 @@ void help()
     std::cout << "patchmanager" << std::endl;
     std::cout << std::endl;
     std::cout << "Usage:" << std::endl;
-    std::cout << "  patchmanager                : Run Patchmanager as daemon" << std::endl;
+    std::cout << "  patchmanager [--help]       : Print this help text" << std::endl;
     std::cout << "  patchmanager -a <Patch>     : Enable and activate a Patch" << std::endl;
     std::cout << "  patchmanager -u <Patch>     : Deactivate and disable (unapply) a Patch" << std::endl;
     std::cout << "  patchmanager --unapply-all  : Deactivate and disable (unapply) all Patches" << std::endl;
-    std::cout << "  patchmanager --daemon       : Daemonize Patchmanager" << std::endl;
+    std::cout << "  patchmanager --daemon       : Start Patchmanager as daemon" << std::endl;
 }
 
 int main(int argc, char **argv)
--- a/src/bin/patchmanager-daemon/main.cpp
+++ b/src/bin/patchmanager-daemon/main.cpp
@@ -1,8 +1,13 @@
 /*
  * Copyright (C) 2013 Lucien XU <sfietkonstantin@free.fr>
  * Copyright (C) 2016 Andrey Kozhevnikov <coderusinbox@gmail.com>
+ * Copyright (c) 2021, Patchmanager for SailfishOS contributors:
+ *                  - olf "Olf0" <https://github.com/Olf0>
+ *                  - Peter G. "nephros" <sailfish@nephros.org>
+ *                  - Vlad G. "b100dian" <https://github.com/b100dian>
  *
- * You may use this file under the terms of the BSD license as follows:
+ * You may use this file under the terms of the 3-clause BSD license,
+ * as follows:
  *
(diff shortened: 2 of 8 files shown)


Olf0:
Open question: What is the technical difference between calling 'patchmanager` (old: "run as daemon, new: "Run Patchmanager as daemon")](https://github.com/sailfishos-patches/patchmanager/pull/337/files#diff-81d57483ccc1832b5475dbd44aa2d3564e809494f3f077ddad12c0027fd720deR56) and calling 'patchmanager` (old: "daemonize", new: "Daemonize Patchmanager")](https://github.com/sailfishos-patches/patchmanager/pull/337/files#diff-81d57483ccc1832b5475dbd44aa2d3564e809494f3f077ddad12c0027fd720deR60) and how to express / differentiate that concisely in these two statements?  Currently they sound confusingly similar.

nephros:
> 1. What is the technical difference between calling `patchmanager` and calling `patchmanager --daemon` 

`patchmanager --daemon` is how it is launched from the d-bus service so it backgrounds and starts to accept calls via that service.

Calling `patchmanager` actually does not do anything, it just checks for the root user and refuses to run if it isn't root, otherwise it will just output the help info.

https://github.com/sailfishos-patches/patchmanager/blob/master/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L941

Olf0:
> > 1. What is the technical difference between calling `patchmanager` and calling `patchmanager --daemon`
> 
> `patchmanager --daemon` is how it is launched from the d-bus service so it backgrounds and starts to accept calls via that service.
> 
> Calling `patchmanager` actually does not do anything, it just checks for the root user and refuses to run if it isn't root, otherwise it will just output the help info.
> 
> Therefore I conclude that the first line in the usage output is plain wrong.
> 
> https://github.com/sailfishos-patches/patchmanager/blob/master/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L941

Thanks, this is what I failed to find quickly yesterday night.

> Calling `patchmanager` actually does not do anything, it just checks for the root user and refuses to run if it isn't root, otherwise it will just output the help info.

[Oh, obviously](https://github.com/sailfishos-patches/patchmanager/blob/master/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L972).

> Therefore I conclude that the first line in the usage output is plain wrong.

So I will change it to "Print this help text.", O.K.?

I also think that there should be an option `--help`, which achieves the same.  Currently any unknown argument results in this being sent to the daemon vis D-Bus, printing [D] unknown:0 - void PatchmanagerObject::process() Has arguments, sending D-Bus message and quit.`

Three more things I am considering:
1. I never called `patchmanager-daemon [-a|-u|--unapply-all]`, always used `patchmanager [-a|-u|--unapply-all]`.  AFAICS (and tried) this is wrong.  Hence shall drop the `-daemon` suffix, correct?
2. Document the option `--reset-system` here.  "Fully reset Patchmanager" would be my first take, unless you can easily think of a better or more specific phrase.  I know you once researched what `--reset-system`, if you do not remember, never mind, IMO a very abstract phrase is better than nothing.
3. Conceptually, I always modelled the `patchmanager` command as a CLI front-end for the `patchmanager-daemon`.  I plan to denote that above all options, where it currently solely states `patchmanager-daemon`, which does not make much sense IMO.

Olf0:
Oh, [line 943](https://github.com/sailfishos-patches/patchmanager/blame/master/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L943) should be 
`} else if (args.count() == 2 && args[1] == QStringLiteral("--reset-system")) {`
, right?

nephros (approved):

nephros (approved):

nephros:
On the topic of `--reset-system`, I'd much prefer if it stayed undocumented.

It causes a rather radical noninteractive change, (it determines the rpms owning any files changed by patches, download them and force-installs them).

As such it's a legacy from the days before the ld-preload method and unlikely to be needed any more.

nephros:
> Three more things I am considering:
> 
> 1. I never called `patchmanager-daemon [-a|-u|--unapply-all]`, always used `patchmanager [-a|-u|--unapply-all]`.  AFAICS (and tried) calling `patchmanager-daemon` is wrong and yields nothing.  Hence the `-daemon` suffix shall be dropped throughout the help text, correct?
> [...]
> 
> 3. Conceptually, I always modelled the `patchmanager` command as a CLI front-end for the `patchmanager-daemon`.  I plan to denote that above all options, where it currently solely states `patchmanager-daemon`, which does not make much sense IMO.

"There is no Dana. There is only Zuul".

`patchmanager-daemon` does not exist as a binary. It's also not used anywhere in the output AFAICS. Not sure where that is coming from. There is only the source directory called that.  
The `patchmanager` executable can operate in  both (dbus-) daemon and client mode (when called with any option other than `--daemon`).

nephros:
> Even when in client mode, it uses a running daemon instance to actually do anything (triggered by a dbus message).

Heh *except* for `--reset-system` which does not go through dbus.

Olf0:
> […] Even when in client mode, it uses a running daemon instance to actually do anything (triggered by a dbus message).

This is what confused me into guessing that there must be a separate daemon when reading tĥe help text.

> > Even when in client mode, it uses a running daemon instance to actually do anything (triggered by a dbus message).

> Heh *except* for `--reset-system` which does not go through dbus.

… as the [lines 941-947](https://github.com/sailfishos-patches/patchmanager/blob/master/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L941-L947) / [my code-snippet]https://github.com/sailfishos-patches/patchmanager/pull/337#issuecomment-1381981308) show.

Does your approval also cover this code-snippet?  I.e., does it look O.K.?

Olf0:
> There is no ~~Dana~~  `patchmanager-daemon`, there is only ~~Zuul~~ `patchmanager`.

That did ring a bell, but I had no clue which, so I looked it up: O.K., that was in 1984!
Did not know though, that this phrase has become a meme.  IIRC it is one of the truly cool scenes in this movie, but much of it is "Klamauk" (I guess "a whimsy" comes close, but is not really an equivalent).

nephros:
> … as the [lines 941-947](https://github.com/sailfishos-patches/patchmanager/blob/master/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L941-L947) and [my code-snippet](https://github.com/sailfishos-patches/patchmanager/pull/337#issuecomment-1381981308) show.
> 
> Does your approval also cover this code-snippet? I.e., does it look O.K.?

Yep, LGTM.

> … as the [lines 941-947](https://github.com/sailfishos-patches/patchmanager/blob/master/src/bin/patchmanager-daemon/patchmanagerobject.cpp#L941-L947) and [my code-snippet](https://github.com/sailfishos-patches/patchmanager/pull/337#issuecomment-1381981308) show.
> 
> Does your approval also cover this code-snippet? I.e., does it look O.K.?

Yep, LGTM.

PR #359, opened 2023-01-16
fix systemctl location

nephros:
Not sure in which SFOS release `systemctl-user` moved from `/bin/` to `/usr/bin`, but on 4.4.0.72 that's where it lives.
Do we need to take care about this location in any of the older supported versions?

Also removes the double execution of it which I don't think was on purpose.

### DIFF
--- a/src/bin/patchmanager-daemon/patchmanagerobject.cpp
+++ b/src/bin/patchmanager-daemon/patchmanagerobject.cpp
@@ -116,7 +116,7 @@ static const QString PM_UNAPPLY = QStringLiteral("/usr/libexec/pm_unapply");
 static const QString BIN_UNZIP = QStringLiteral("/usr/bin/unzip");
 static const QString BIN_TAR = QStringLiteral("/bin/tar");
 static const QString BIN_PKCON = QStringLiteral("/usr/bin/pkcon");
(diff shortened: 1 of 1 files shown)


Olf0:
> Not sure in which SFOS release `systemctl-user` moved from `/bin/` to `/usr/bin`, but on 4.4.0.72 that's where it lives. Do we need to take care about this location in any of the older supported versions?
> 
> Also removes the double execution of it which I don't think was on purpose.



> Not sure in which SFOS release `systemctl-user` moved from `/bin/` to `/usr/bin`, but on 4.4.0.72 that's where it lives. Do we need to take care about this location in any of the older supported versions?

Yes, IIRC all.  `systemctl` and `systemctl-user` were moved from `/bin` to `/usr/bin` with the release of SFOS 3.4.0 or 3.3.0 (hence all versions PM ≥ 3.1 supports).  Will look that up, just to be sure.

> Also removes the double execution of it which I don't think was on purpose.

:+1:

Olf0:
> `systemctl` and `systemctl-user` were moved from `/bin` to `/usr/bin` with the release of SFOS 3.4.0 …

… see https://forum.sailfishos.org/t/release-notes-pallas-yllastunturi-3-4-0/2258#toc-h2-technical-changes

Olf0 (approved):
Fine!

PR #373, opened 2023-01-17
Update translation string for #372

nephros:
I should have pushed this to the feature branch for #372 before merging.
Oh well, too late.


Olf0 (approved):
I like the branch name `transifix`!

Yes, a little detour; never mind.

PR #387, opened 2023-01-19
"click" → "tap"

Olf0:
"click"→"tap" in the source code file and also in the source strings TS file to avoid a `lupdate` / `lrelease` run.

### DIFF
--- a/src/qml/WebPatchPage.qml
+++ b/src/qml/WebPatchPage.qml
@@ -404,10 +404,10 @@ Page {
                                 color: Theme.highlightColor
                                 text: fileDelegate.isInstalled
                                     ? ( fileDelegate.isReinstallable
-                                        ? qsTranslate("", "[click to re-install]")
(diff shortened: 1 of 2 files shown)


nephros (approved):
LGTM.

When translating, we should strive to keep it short, space is limited where this is used.
The current German "zum (neu) Installieren tippen" *could* be a bit too long. "tippe zum..." would at least save two characters :)

Olf0:
> When translating, we should strive to keep it short, space is limited where this is used. The current German "zum (neu) Installieren tippen" _could_ be a bit too long.

What about "[Tippe zum Installieren]"?
<sub>Mmmh, my gut feeling for grammar suggest to rather use "[Tippe, um zu installieren]", but never mind: my grammar feeling errs regularly, so pick whatever you deem as appropriate.</sub>

Olf0:
Also "Antippen" came to my mind: 

[Antippen zum Installieren]
*or*
[Antippen, um zu installieren]

PR #393, opened 2023-01-23
Extend README a bit

Olf0:


Olf0 (commented):
Should be O.K. now.

Olf0:
> @nephros, if [`tests/README_patching.md`](https://github.com/sailfishos-patches/patchmanager/blob/master/tests/README_patching.md) is simply called `tests/README.md` it should be automatically displayed when viewing https://github.com/sailfishos-patches/patchmanager/blob/master/tests
> 
> You might consider to change that (or tell me, if you want that). The the link in the top README would have to be adapted, then. Hence it might make sense to add this change to this PR, if you want that.

I interpret no-reply, while you have approved this PR, as a "no, at least not in this PR", correct?

Olf0:
> Thanks for the hint, done! :)

[Works fine](https://github.com/sailfishos-patches/patchmanager/tree/master/tests#readme), is looking good IMO.

Issue #327, opened 2023-01-05
`tar_git` at SFOS-OBS assembles an incorrect changelog


Olf0:
**PATCHMANAGER VERSION**: 3.x.y

#### BUG DESCRIPTION

`tar_git` assembles an incorrect changelog, see [e.g. for v3.2.4](https://build.merproject.org/package/view_file/sailfishos:chum:testing/patchmanager/_service:tar_git:patchmanager.changes?expand=1).

Consequently `rpmlint` at the SFOS-OBS complains:
`patchmanager.aarch64: W: incoherent-version-in-changelog 3.2.3 ['3.2.4-1.12.1.jolla', '3.2.4']`
`The latest entry in %changelog contains a version identifier that is not coherent with the epoch:version-release tuple of the package.`

AFAICS, Patchmanager currently does not provide an `rpm/patchmanager.changes` file, a `%changelog` section in the spec file or any other file accumulating the changes for each release.

#### STEPS TO REPRODUCE

Compile Patchmanager 3 at the SailfishOS-OBS.

#### ADDITIONAL INFORMATION

As we provide proper changelogs in [Patchmanager's releases page at GitHub](https://github.com/sailfishos-patches/patchmanager/releases) and git provides full traceability on commit level, IMO `git_tar`'s flaw provides insufficient reason to address this by duplicating administrative work (i.e., maintain two changelogs, which must be kept in sync).

Consequently I labelled this as `#wontfix`.  If someone thinks this is wrong, please remove the `#wontfix` label and address this issue.

nephros:
That rpmlint complaint stems from the Version-Release mangling the OBS build process does.

Even if there were a .changelog file this would appear for every rebuild (the x.y.z-jolla revisions OBS generates).

AFAIK one could use a .changes.in file which can generate a changelog from git, and possibly would shut up rpmlint.

Olf0:
> That rpmlint complaint stems from the Version-Release mangling the OBS build process does.

Yes, see [my analysis of Jolla's `tar_git`](https://github.com/sailfishos-chum/main/issues/77#issuecomment-1079780452).

> Even if there were a .changelog file this would appear for every rebuild (the x.y.z-jolla revisions OBS generates).

Ack. 

> AFAIK one could use a .changes.in file which can generate a changelog from git, and possibly would shut up rpmlint.

IMO a lot of effort and maintenance for minimal, and only potential gain (it may not work this way, too).

[...]

PR #395, opened 2023-01-23
[spec file] Insert dummy changelog with a correct entry 

Olf0:
… to calm down some versions / configurations of `rpmlint`: Missing `%changelog` section or file is punished with an error, an inconsistent version number only with a warning.

- Update homepages to OpenRepos pages (looks more representative)

- Post release version increase

### DIFF
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -291,5 +291,5 @@ exit 0
 %{_datadir}/themes/%{theme}/meegotouch/z2.0/icons/*.png
 
 %changelog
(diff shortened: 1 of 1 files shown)


Olf0:
> currently, `rpm -q --changelog patchmanager` does show a nice changelog generated from git.
> 
> Will it still have that after this change?

Likely not.  You pointed at the mechanism which seems to go wrong at the SailfishOS-OBS (somehow extracting the changelog), which I wanted to research for some time.

But as other things do have higher priorities, I am deferring this topic and setting this PR back into draft status.

TBE (to be evaluated) more thoroughly.

Olf0:
See https://github.com/sailfishos-patches/patchmanager/issues/327#issuecomment-1404304969
⇒ Back into "awaiting review" state

Issue #405, opened 2023-01-29
[Bug] allow incompatible not working as expected


nephros:
it seems #334 introduces one or several bugs where

1. the setting isn't initialized properly after updating from a previous version of PM. One has to toggle the option list to make the wanted option stick.
2. ~~while incompatible patches are allowed to be activated and show up as such, the aren't actually applied by the PM backend.~~ <-- not actually *this* bug here

test example patches: 
 - "smooth remorse timers"
 - "browser tab pulley menu"

Olf0:
Do I understand correctly, that point 2 is not a regression, i.e., "always has been this way for PM3"?

Does "the setting" in point 1 address only a Patch's "enabled / disabled" status as GUI (= user) setting?

Olf0:
> The interesting part I'm not sure about is:
> If I
>  * LD_PRELOAD=patchmanager.so[version X] lipstick
>  * ... do things
>  * update PM so it installs patchmanager.so[version Y] 
>  * do _not_ restart lipstick
> 
> ... will patchmanager.so[version Y] do its thing? I think not, as it's not really loaded into the lipstick process.

Soi it is basic requirement would be to reload the shared library `patchmanager.so` (in its freshly deployed version Y), which means to restart the daemonized Patchmanager running in the background (correctly understood?): Is this feasible at all?

Olf0:
> About 1, I think what happens is:
> 
> […]
> now at this point (on a never-migrated system), `PatchManager.sfosVersionCheck` is undefined, so it gets set to `VersionCheck.Strict`. So far so good, we agreed that we should default to the conservative option.
> 
> But that I think does _not_ trigger the `onCurrentIndexChanged` signal handler, so the actual PM setting never gets set.
> 
> Only interactively toggling the switch will do it.

Reading this, it sound not hard to resolve, if that is indeed the cause, right?

Olf0:
> […] (At most, we can give a note to the user they should reboot after install but, _meh_.)

Well, this is better than something not working without such a indication ("without a reboot strange things may happen").  Jolla's `version dup` does exactly that (though it is a CLI app).

> BUT, in reality this should rarely be needed as breaking changes to the .so almost never happen.

Mmmh, *you* observe it, so anybody else could.
I will follow any assessment from you here, because I do not have the time and barely have the know-how to retrace that, but just want to ask "checks & balances" questions.

[...]

PR #406, opened 2023-01-29
Split developermode: move migration to initial page

nephros:
Fixes: #405


Olf0 (commented):
Mostly clueless comment: LGTM, but it is QML.
I.e., this is for you to approve @b100dian. :smiley:

b100dian (approved):
Well this is just moving it to the Settings app -> Patchmanager page (instead of Settings app, Patchmanager, Settings) so it should be safe

PR #407, opened 2023-01-29
Add version info output

nephros:
A simple one.

Add version info to console (help) output. Just because sometimes it's useful to check.

### DIFF
--- a/src/bin/patchmanager-daemon/main.cpp
+++ b/src/bin/patchmanager-daemon/main.cpp
@@ -79,7 +79,7 @@ int main(int argc, char **argv)
         return 0;
     }
 
-    app.setApplicationVersion(BUILD_VERSION);
+    app.setApplicationVersion(QStringLiteral(BUILD_VERSION));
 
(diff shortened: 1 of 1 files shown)


Olf0:
So calling [`QStringLiteral`](https://github.com/sailfishos-patches/patchmanager/pull/407/files#diff-81d57483ccc1832b5475dbd44aa2d3564e809494f3f077ddad12c0027fd720deL83) was superfluous?

Olf0 (approved):
I hope my change of wording makes sense: I always avoid a "v" indicating a version number, if if directly follows the product name.  E.g., Patchmanager 3 instead of Patchmanager v3.  In also think it makes sense to capitalise Patchmanager in the header, where the term is a title and product name; in contrast to the body, where it has to be spelled as it is called at the command line. 

Approving, but posed a single checks and balance question.  You may simply answer "Yes" and merge, or "No" and alter the code.

Olf0 (commented):
Now this is fully LGTM.

PR #408, opened 2023-02-01
PatchPage: move to patchDevelMode

nephros:
more fallout from my previous work :/

makes the patch log and legacy format displays use the new property


Olf0 (approved):
Does look good and is so trivial, that I have no issue at all to approve a change in a QML source file.

PR #410, opened 2023-02-03
[Fix] Don't care about icon scaling

nephros:
SFOS 4.5 introduces z2.5, lets just  include any scale


Olf0:
AFAICS, the `3.2.6-fixes` branch contained only the icon scaling fix for SFOS 4.5.0: So I assume it can be deleted by hitting the `Delete brach` button above, right?

PR #411, opened 2023-02-03
Indicate presence of a Settings page

nephros:
some patches come with a settings page.

users often are not aware, so lets add a hint

PR #416, opened 2023-02-06
Remove tinting, use Theme error color for compat display

nephros:
See: https://github.com/sailfishos-patches/patchmanager/pull/411#discussion_r1096832591

Tinting never did anything but apply the color `"red"`, because `Qt.tint()` wants the second color to be more or less transparent, which `"red"` is not.

`Theme.errorColor` is not as bright and saturated as `"red"` but it matches the system theme.


b100dian (approved):
Nice, thanks!

PR #417, opened 2023-02-11
[CI workflow] Update LATEST to 4.5.0.16 and more

Olf0:


Olf0:
> However, are you aware of this?

Faintly; i.e., reading this I remember the fact, and logically that this will break something sooner or later, but I was not the exact consequences: With binaries built for which release.

Hence, thank you for bringing this up.

> (I'm aware that this here is about the latest release, but it may be useful to make a note for older build versions.)

Nah, here I rather prefer a technical solution, as it seems to be easily feasible.

What are the binaries build via CI runs intended for?
* Primarily to see, if they build correctly.  For this purpose all this does not matter.  This is why the CI run for pulls to the master branch build against the oldest supported release (3.4.0) for `armv7hl` only (smallest docker image to download and only a single architecture to minimise CI workflow run time) and the CI run on tag creation builds against the oldest supported releases (3.4.0 for `armv7hl` and `i486`, plus 4.0.1 for `aaarch64`) and the "latest" for all three architectures (`aarch64`, `armv7hl` and `i486`).
* But I do append these builds to each release, so users may install it.
* At OpenRepos I publish the builds for "latest", because these for "oldest supported" do not install on recent SFOS (4.4.0, already).

> > To make it clear, rpms created with the 4.5.0.16 toolchain can no longer be installed on e.g. a Jolla 1 phone, because the rpm version in SFOS 4.3 cannot handle zstd compression it seems.

So, what do you suggest?  Jolla phone users can use the build for "oldest supported" and the proper solution is IMO SailfishOS:Chum by providing builds for each release (as long as Jolla still maintains the SFOS-OBS), but adding another SFOS release to the "CI on tag creation" is fine.  Which one out of {4.1.0,4.2.0,4.3.0,4.4.0}?

Olf0:
> > To make it clear, rpms created with the 4.5.0.16 toolchain can no longer be installed on e.g. a Jolla 1 phone, because the rpm version in SFOS 4.3 cannot handle zstd compression it seems.

BTW, they cannot any way, due to other breaking changes.  But please see [my research posted at FSO](https://forum.sailfishos.org/t/q-4-5-will-switch-rpm-compression-what-does-that-mean-for-existing-rpms/13153/18?u=olf) and [query for more checks](https://forum.sailfishos.org/t/q-4-5-will-switch-rpm-compression-what-does-that-mean-for-existing-rpms/13153/19?u=olf), for which RPMs built by Coderus' docker images with the Sailfish-SDK (e.g., here at GitHub by CI workflows; [Storeman provides a larger variety of builds](https://github.com/storeman-developers/harbour-storeman/releases)) or at the SailfishOS-OBS (e.g., from the SailfishOS:Chum repository) should work well.

Olf0:
> > and who knows which hardware GitHub really uses.
> 
> AFAIK that's not really relevant, because all SFOS SDK tooling is translated via qemu to the target arch, and that qemu is always amd64. So the build host will always have to be `amd64` 

Sounds reasonable, but …

> - and yes, i486 targets usually compile faster than the others (and I guess that's also because of qemu).

… I have not observed this: Sometimes I watch the CI workflow's log-output "live" to spot things (reminds one of a Linux compilation on an real i486, well with v2.x.y; modern kernel sources would take ages to compile).  But, as it is definitely not slower, I will switch to `i486` for the runs on "pulls".

> I have never inspected GH in detail but if it's anything like GitLab, then the layers are `[hardware host] -> ( [likely another layer like Amazon or Azure ] ) -> [ ci runner docker  ] -> [ coderus docker ] -> [ qemu-translated cross-compiling environment ]`

At GitHub I do not know anything about the first to levels you denoted for Gitlab.com, because that is opaque.  What one can choose from is a {Ubuntu,MacOS,Windows} in a VM (I actually prefer "on" an virtual machine).  On Ubuntu (likely on MacOS and Widows, too) `dockerd` is already running (= pre-installed), though one can easily install anything via (Ubuntu-style) `sudo apt-get install …`.  Coderus' Sailfish-SDK docker image is started with `docker run …`, which pulls it from Docker's default image repo at `docker.io` if not already pulled by a earlier `docker pull|run` call in that VM instance.  And yes, within the docker "container" `sb2` (Scratchbox 2) runs `quemu` in a `chroot` environment.  So there is at least one opaque full virtualisation layer underneath the Ubuntu, plus the docker-based container in which Coderus' Sailfish-SDK image is started, in / on which the `chroot-quemu` is controlled by `mb2` (the outside tool for controlling the "inside") and the `sb2` inside of it.

> so in the worst case there are going to be several arch conversions down to the next layer, only one of which we can control.

Rather 2½, WRT "control", but there is no way one can chose something different.

> (BTW, that `[ ci runner docker] -> [ coderus docker ]` double setup is why you can't use any of the supplied caching on Gitlab, but it's not obvious.)

On GitHub it is that their `cache` action requires one to point to a user accessible directory or file(s) to cache.  Currently `dockerd` runs in the root contect and downloads docker images to some non-user-accessible location on the filesystem: Running Docker "rootless" is the way out, but I am busy with oh so many detours.  Still hope to return to that task some time.

Olf0:
@nephros, now it has become a bit more than the original title indicates: If you want to re-review, I will gladly consider any new comments.

PR #420, opened 2023-02-24
compat check: fix logic for Web Patch (re)install

nephros:
Some fallout from #408 

The logic prevents installation of incompatible patches from Web Catalog despite (or in the reverse meaning of) the VersionCheck setting.

At first I attributed this to the migrated settings bug #405, but it's a separate issue.

Please review the logic, could be I missed a case.

### DIFF
--- a/src/qml/WebPatchPage.qml
+++ b/src/qml/WebPatchPage.qml
@@ -342,12 +342,13 @@ Page {
                     property bool isInstalled: !!container.versions && container.versions[modelData.project] == modelData.version
                     property bool isCompatible: (modelData.compatible.indexOf(PatchManager.osVersion) >= 0)
                     property bool forceCompatible: PatchManager.sfosVersionCheck !== VersionCheck.Strict
-                    property bool isReinstallable: isInstalled && (isCompatible || forceCompatible)
+                    property bool isInstallable: isCompatible || forceCompatible
+                    property bool isReinstallable: isInstalled && isInstallable
 
                     onClicked: {
                         if (isReinstallable) {
                             remorseAction(qsTranslate("", "Re-Install Patch %1").arg(patchData.display_name), installPatch)
-                        } else if (!isCompatible && !forceCompatible) {
+                        } else if (!isInstallable) {
                             errorMessageComponent.createObject(fileDelegate, {text: qsTranslate("", "This Patch is incompatible with the installed SailfishOS version.")})
                         } else {
                             remorseAction(qsTranslate("", "Install Patch %1").arg(patchData.display_name), installPatch)


src/qml/WebPatchPage.qml
```diff
@@ -340,16 +340,17 @@ Page {
...
                     property bool isInstalled: !!container.versions && container.versions[modelData.project] == modelData.version
-                    property bool isCompatible: modelData.compatible.indexOf(PatchManager.osVersion) >= 0
-                    property bool isReinstallable: isInstalled && isCompatible
+                    property bool isCompatible: (modelData.compatible.indexOf(PatchManager.osVersion) >= 0)
+                    property bool forceCompatible: PatchManager.sfosVersionCheck !== VersionCheck.Strict
+                    property bool isReinstallable: isInstalled && (isCompatible || forceCompatible)
 
                     onClicked: {
-                        if ((PatchManager.sfosVersionCheck !== VersionCheck.Strict) && !isCompatible) {
+                        if (fileDelegate.isReinstallable) {
```
b100dian, line 348: consistency nit: is `fileDelegate.` needed
-  here specifically
- in all accesses for the properties changed in this PR (`isCompatible`, `forceCompatible`) or
-  in none of the cases?

src/qml/WebPatchPage.qml
```diff
@@ -340,16 +340,17 @@ Page {
...
+                            remorseAction(qsTranslate("", "Re-Install Patch %1").arg(patchData.display_name), installPatch)
+                        } else if (!isCompatible && !forceCompatible) {
```
b100dian, line 350: this is equivalent to `!(isCompatible || forceCompatible)` which, when true - and because `isReinstallable` was false because we're on the else branch - tells us nothing about `isInstalled`.
However when `!(isCompatible || forceCompatible)` is false (the next else) this does tells us that `isCompatible || forceCompatible)` is true which tells us that `isInstalled` was the one false failing the `isReinstallable check`.

Therefore re-reading the whole if/else is like:
```
if (isReinstallable) {
  // re-install
} else if (!isCompatible && !forceCompatible) {
  // error 
} else since (isCompatible || forceCompatible) {
  // install
}
```
which looks sound to me.

Before, I think the "only" error was forceCompatible which not tested correctly (it needed a !) 
```
if (forceCompatible && !isCompatible) {
  // error - oops, we are erroring out if forceCompatible is on?
} else since (!forceCompatible || isCompatible) if (!isInstalled) {
  // install - oops we might install if forceCompatible is off?
} else since (!forceCompatible || isCompatible) && isInstalled {
  // re-install we might re-install if forceCompatible is off?l 
}
```

Thanks for making this more readable :)

b100dian (approved):

PR #425, opened 2023-04-13
Set vendor statically to `chum` for all builds …

Olf0:
…, regardless where they are built, to avoid blocked upgrade paths (as, e.g., observed in [Storeman issue \#406](https://github.com/storeman-developers/harbour-storeman/issues/406#issuecomment-1503913989)) due to [vendor stickiness](https://en.opensuse.org/SDB:Vendor_change_update).


Olf0:
> I'm not 100% convinced that is the "formally" correct fix for the situation, but it's the pragmatic way with least pain for users, …

I fully concur: I think & feel the same way about this, but know no better way to handle it ATM.

PR #430, opened 2023-04-15
Add three new Testcases: Version Checks

nephros:
These allow playing around with version check strictness.

### DIFF
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -28,6 +28,7 @@ Release:    1
 Group:      Qt/Qt
 License:    BSD-3-Clause
 URL:        https://github.com/sailfishos-patches/%{name}
+Vendor:     chum
 Source0:    %{url}/archive/%{version}/%{name}-%{version}.tar.gz
 # Note that the rpmlintrc file MUST be named exactly so according to
 # https://en.opensuse.org/openSUSE:Packaging_checks#Building_Packages_in_spite_of_errors


Olf0 (commented):
This is LGTM, but I would also like to have @b100dian taking a look at it.  If he cannot spare the time within a week or so, I will approve to avoid stalling these test cases.

b100dian:
Thanks for the direct pings @Olf0 . You guys are doing a great thing with this project and I sometimes have the impression everything is so smooth I don't need to do a thing ;) Glad I can still help here and there though!

@nephros Please remind me, how does the test app test the negative "PM should not allow to activate or install it" ? Thanks!

Olf0:
> I probably should add something to the app as well, that's right.

@nephros, are you planning to do that in this PR or in a new one?

Depending on your reply I would (suggest to) ping Vlad once again, either now or when the test app was expanded.

Olf0:
@b100dian: Please see the two messages above.

Olf0:
> You guys are doing a great thing with this project and I sometimes have the impression everything is so smooth I don't need to do a thing ;)

BTW,
> [Personally I will try to finish a couple of open tasks in the next months, before seriously considering to try new [non-SailfishOS] things.](https://github.com/sailfishos-chum/main/pull/91#issuecomment-1455154073)

I.e., I will slowly wind down my SFOS-related activities.  Long story with background info is linked to by the quote above.

b100dian (approved):

PR #432, opened 2023-04-19
Trivial (whitespace) changes

nephros:
- do some whitespace alignments
 - reorder constants so they are grouped a bit more logically
 - also, rename one constant

Mainly to improve readability for me.


Olf0 (approved):
LGTM, literally!  I.e., I just quickly glanced over these changes, but did not check, e.g., that these are the only (two) occurrences of ´PATCH_METADATA_FILE´.

PR #433, opened 2023-05-01
Add comment WRT values for the Group field

Olf0:
Never saw a reference, so document it now.

### DIFF
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -25,6 +25,8 @@ Name:       patchmanager
 Summary:    Allows to manage Patches for SailfishOS
 Version:    3.2.7
 Release:    1
+# The Group field should comprise one of the groups listed here:
+# https://github.com/mer-tools/spectacle/blob/master/data/GROUPS
 Group:      Qt/Qt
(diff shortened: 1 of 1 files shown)


Olf0:
> The `Group` tag is mandatory if the spectacle tool is used via a .yaml file, as spectacle wants it to be filled with a value from that list.

Ack.

> As we don't use spectacle (and `Group` is optional from an RPM point of view) we might as well remove that Tag completely. (Qt/Qt isn't a very valuable entry anyway.)

O.K.
I simply intended to document where valid values can be found, so one can pick an more appropriate `Group` tag later.

I have no idea if anything else but Spectacle evaluates the `Group` tag; if you are sure that nothing else does, I am fine with removing it.

P.S.: The `Group: Qt/Qt` statement seems to be something which was copied to almost all SFOS applications.

Olf0:
Looking at the list, I think that both `Applications/System` and `System/System Control` (but nothing else) suits Patchmanager well.  I picked the latter, because it is less generic (as `Qt/Qt` was criticised as too generic, which is justified).

PR #436, opened 2023-05-26
multi-apply-tool: fix some syntax errors

nephros:
It seems the history of #302 ended up in some merge/syntax errors

These are some quick changes to make the script work again.

PR #437, opened 2023-05-26
Implement a Last-Known-Good feature

nephros:
Lets improve usability in case of failure some more: make it easier to recover.
Contributes-To: #277

Currently:

 - saves the list of applied patches each time "autoapply" succeeds, i.e. at boot or Lipstick launch
 - list is saved as new entry in the normal .ini file , `/etc/patchmanager2.conf`
 - if PM goes into failure mode, adds the user menu option to load that list
 - after this, 'resolve failure' should have a valid list of active patches

adaed74a9be0 adds parameters to patchmanager-tool, which is currently broken, so:
Depends-on: #436

### DIFF
--- a/src/bin/patchmanager-daemon/dbus/org.SfietKonstantin.patchmanager.xml
+++ b/src/bin/patchmanager-daemon/dbus/org.SfietKonstantin.patchmanager.xml
@@ -107,10 +107,10 @@
         <method name="getLoaded">
             <arg name="loaded" type="b" direction="out" />
         </method>
-        <method name="revertToLastGood">
+        <method name="restorePatchList">
             <annotation name="org.freedesktop.DBus.Method.NoReply" value="true"/>
         </method>
-        <method name="setKnownGood">
+        <method name="backupWorkingPatchList">
             <annotation name="org.freedesktop.DBus.Method.NoReply" value="true"/>
         </method>
         <method name="resolveFailure">
--- a/src/bin/patchmanager-daemon/main.cpp
+++ b/src/bin/patchmanager-daemon/main.cpp
@@ -61,8 +61,8 @@ void help()
     std::cout << "  patchmanager -a <Patch>        : Enable and activate a Patch" << std::endl;
     std::cout << "  patchmanager -u <Patch>        : Deactivate and disable (unapply) a Patch" << std::endl;
     std::cout << "  patchmanager --unapply-all     : Deactivate and disable (unapply) all Patches" << std::endl;
-    std::cout << "  patchmanager --save-as-good    : Save list of enabled Patches as 'known good'" << std::endl;
-    std::cout << "  patchmanager --load-known-good : Enable 'known good' Patches" << std::endl;
+    std::cout << "  patchmanager --backup-working  : Save list of enabled Patches as \"working\"" << std::endl;
+    std::cout << "  patchmanager --restore-working : Enable backup list of \"working\" Patches" << std::endl;
     std::cout << "  patchmanager --daemon          : Start Patchmanager as daemon" << std::endl;
 }
 
--- a/src/bin/patchmanager-daemon/patchmanagerobject.cpp
+++ b/src/bin/patchmanager-daemon/patchmanagerobject.cpp
@@ -151,14 +151,14 @@ static const QString KEYBOARD_CODE    = QStringLiteral("keyboard");
   \class PatchManagerObject
   \inmodule PatchManagerDaemon
 
-  \brief The Patchmanager Daemon.
+  \brief The Patchmanager daemon.
 
   A D-Bus activated background service which manages patch un/installation,
   listing, de/actvation, and communication with the preload library.
 
-  PatchManager is usually launched by its DBus service.
+  Patchmanager is usually launched by its D-Bus service.
   The binary can also serve as a simple command-line client to a running
-  daemon. See the output of \c{patchmanager --help} for more information.
+  daemon.  See the output of \c{patchmanager --help} for more information.
 
 */
 
@@ -178,7 +178,7 @@ static const QString KEYBOARD_CODE    = QStringLiteral("keyboard");
         \value NotifyActionFailedUnapply
             unapplying was not successful
         \value NotifyActionUpdateAvailable
-            one of the patches has an update
+            one of the Patches has an update
 */
 
 QString getLang()
@@ -259,7 +259,7 @@ bool PatchManagerObject::makePatch(const QDir &root, const QString &patchPath, Q
 }
 
 /*!
-    Sends a notification to the user. \a patch is the patch name, \a action one of:
+    Sends a notification to the user.  \a patch is the Patch name, \a action one of:
     \sa PatchManagerObject::NotifyAction
 */
 void PatchManagerObject::notify(const QString &patch, NotifyAction action)
@@ -342,7 +342,7 @@ void PatchManagerObject::notify(const QString &patch, NotifyAction action)
 }
 
 /*!
-    Returns the list of applied patches via getSettings().
+    Returns the list of applied \a patches via \c getSettings().
     \sa putSettings(), setAppliedPatches()
 */
 QSet<QString> PatchManagerObject::getAppliedPatches() const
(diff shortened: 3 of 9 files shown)


b100dian:
Changes look sound, I have yet to test them though

Olf0:
> @Olf0 is it intended that the build only provides i486 artifacts, or should I raise a bug?

Yes, I the CI run for each commit to a Pull-Request against the `master` branch always was only for a single architecture on the oldest supported SFOS release (currently 3.4.0): It is intended as a simple build check, i.e., not to provide binaries to test.

I recently switched from `armv7hl` to `i486` (upon a suggestion by @nephros), because it compiles faster, likely due to being the native architecture of the build host.  This is also why i picked the oldest supported release: Coderus docker images grown much from each SFOS release to the next release, this way we only pull 2 GB instead of 4 GB from docker hub *for each commit* of a PR to `master`.

Olf0:
> > find a better name for the [menu entry](https://github.com/sailfishos-patches/patchmanager/pull/437/commits/2b04ab65e21893f4eb95847d3ad007573fffe212#diff-da6a1e0621bec30e25650901e8a69669572d302bbccc7f82e5fc81f4b90e4e68R205) - @Olf0, any ideas?
> 
> I am pondering about this. I would like to get rid of the term "known good" in the user-facing dialogues and options, because users may misunderstand its meaning easily.

… misunderstandable in the sense that we as Patchmanager developers *know* that these Patches are *"good"* (however that may be interpreted).  After a while I thought of "O.K. set of Patches" and after translating this to proper lingo ended up with "saved, fine set of Patches" (IOW, "stored, acceptable list of Patches").  I also considered "working" instead of "fine", but "working set" has its own meaning, specifically in IT.

Do you also consider this to be better?  Should I try to alter the strings in this PR accordingly, when I find some time for that?

Olf0:
> Ehm, not happy with the variants.

This is fine: I wanted to do something towards finishing this task, and because I had no good ideas, took two thesauri and played with words.  This is the best I could come up with yesterday, and it appeared artificial to me, both the process and the result.

> Maybe we should just call it a "backup list of patches" or "backup configuration" or so, and in the failure mode offer a "restore from backup"?
> 
> Users should be familiar with these terms, and they are generic enough to change implementation details later.
> 
> Another term that came to mind is "failsafe" instead of known-good.

I fully acknowledge all three statements!

What about "Backup list of working Patches" and "Restore backup list"?

Olf0:
> [...] I intended to contribute with [the wording changes to "Backup list of working Patches" and "Restore backup Patch list"](https://github.com/sailfishos-patches/patchmanager/pull/437#issuecomment-1650825008), plus other terms derived from these two.

Done (and a little more) by https://github.com/nephros/patchmanager/pull/6

After reviewing that PR, you may consider (no need to hurry at all):
> [...] I [prepared a Patchmanager 3.2.10 release](https://github.com/sailfishos-patches/patchmanager/releases), if you, @nephros, plan to finalise this in the upcoming weeks, then I will gladly wait for this PR. Note that I already prepared a changelog line for this PR.

Olf0:
> I feel this feature has not been tested enough to be comfortable packing it into a release just yet.

O.K., I will do a "safe" release without it, then.  But after that I believe we should resolve the diverging feature branches first, by merging this PR and consolidating it and the `more-docs` branch.  While we could do all that in separate development branches, I really would like to have it all merged into `master` for the sake of simplicity: I lost track of which branches in which repository (here and yours) contain new stuff to be merged some day, are experiments or developments which reached a dead end (and hence are stale, but not deleted), or are simple left-overs because they have been merged, but not deleted (the latter, third category presumably dwells only in your repo).  While I also love to keep old experiments around (in case one wants to look at these ideas, again), merged branches ought to be deleted IMO and diverging branches ([my bad here](https://github.com/nephros/patchmanager/pull/6#issuecomment-1676905740)) consolidated into one (preferably `master`).

Olf0:
> P.S. / Edit: [Released `Patchmanager 3.2.10`](https://github.com/sailfishos-patches/patchmanager/releases/tag/3.2.10)

Argl, as usual, with "let us quickly do XYZ, should not pose an issue", … (it does](https://github.com/sailfishos-patches/patchmanager/actions/runs/5858889829/job/15883758907#step:5:300): 
```
patchmanager.cpp: In member function 'void PatchManager::activation(const QString&, const QString&)':
patchmanager.cpp:1020:47: error: unused parameter 'patch' [-Werror=unused-parameter]
 void PatchManager::activation(const QString & patch, const QString & version)
                               ~~~~~~~~~~~~~~~~^~~~~
patchmanager.cpp:1020:70: error: unused parameter 'version' [-Werror=unused-parameter]
 void PatchManager::activation(const QString & patch, const QString & version)
                                                      ~~~~~~~~~~~~~~~~^~~~~~~
patchmanager.cpp: At global scope:
patchmanager.cpp:1028:61: error: declaration of 'void PatchManager::easterReceived(const QString&)' outside of class is not definition [-fpermissive]
 void PatchManager::easterReceived(const QString & easterText);
                                                             ^
patchmanager.cpp:1029:1: error: expected unqualified-id before '{' token
 {
 ^
cc1plus: all warnings being treated as errors
```

Someone wanted "treat warnings, as errors", right? :wink:
But AFAIR, it compiled without any warnings at the point we flipped on this switch, though this is old, unchanged cope, AFAICS.  I will take a look at `git blame`.

b100dian (approved):
> Considering the bumpy history of the "strict checking" feature I'd like this to simmer a bit more.

Shouldn't we just admit that we outsource testing and get through with this? I personally am not in the position to test locally (if no arm64 build comes out) very often.

Olf0:
> Shouldn't we just admit that we outsource testing and get through with this?

This is a conclusion I arrived at in other projects, too.  I just lack the time, so I rather try to improve things than to test thoroughly.  I we want to be "extra careful" we can label a release as beat, but deploy it via the usual update channels (otherwise it will not be broadly tested).

Olf0:
> > Shouldn't we just admit that we outsource testing and get through with this?
> 
> This is a conclusion I arrived at in other projects, too.

I.e. "**yes**".

@nephros, what is your perspective on this?

Olf0:
@nephros, in August 2023 we [had this conversation](https://github.com/sailfishos-patches/patchmanager/pull/437#issuecomment-1677746660) in the comments thread of this PR:
> > I feel this feature has not been tested enough to be comfortable packing it into a release just yet.
> 
> O.K., I […] released [Patchmanager 3.2.11](https://github.com/sailfishos-patches/patchmanager/releases/tag/3.2.11).
> 
> But after that now I believe we should resolve the diverging feature branches first, by merging this PR and consolidating it and the `more-docs` branch.

I just updated [the draft for the Patchmanager 3.2.12 release](https://github.com/sailfishos-patches/patchmanager/releases) and there is definitely sufficient material, new features and time passed (16 months) that we should finalise the 3.2.12 release soon, not Jolla-soon™ :wink::
- Either by creating a release now and then merging this and creating a beta release in the first half of December (which may also include fixes for flaws in Patchmanager 3.2.12).
- Or by merging this now, and creating a release or a beta release out of it.

@b100dian and I [preferred the second choice](https://github.com/sailfishos-patches/patchmanager/pull/437#issuecomment-2099627262) ("broad testing always happens when being deployed to users", which is also very Jolla-style, respectively software industry in general), but I do not really mind as log as we agree on a plan how to proceed with this PR with a timeline ending mid-December latest.

Side note: @nephros, what about merging the `more-docs` branch into `master`?  Do I miss to see a reason why this is a bad idea?

Olf0:
@b100dian, WRT
> I personally am not in the position to test locally (if no arm64 build comes out) very often.

aarch64 builds are generated very easily by setting an arbitrary Git-tag (which should clearly denote that it is a `test` version in its name), which triggers the CI pipeline.

But even easier is to simply hit the `Trigger services` button at [my Patchmanager test-build OBS-repo](https://build.sailfishos.org/package/show/home:olf:patchmanager/patchmanager).

Please note it was really nerve killing to wait for more than 10 Minutes after adding a new commit to a PR to `master` in order to see if it still compiles: Hence the `ci-on-pull_request.yml` configuration only test-builds for i486 on the smallest Sailfish-SDK docker-image feasible (i.e. 3.4.0, which saves gigabytes of downloads from docker-hub and the time these downloads require which is minutes).

Olf0:
@b100dian, I would appreciate much, if you provide some guidance on https://github.com/sailfishos-patches/patchmanager/pull/437#issuecomment-2499303816 when you have time fort that.

PR #442, opened 2023-06-28
Document all the things

nephros:
This is just to let you know that I am in the process of generating inline documentation for all the different PM components.

It started out with me wanting to inline *comment* some code, but ended up with a almost complete qdoc  markup.

This is not finished at all (probably never will be), but it's in a previewable state, hence this draft PR.

Previews of the generated docs can be looked at here: https://htmlpreview.github.io/?https://github.com/nephros/patchmanager/blob/documentation/doc/generated/patchmanager/index.html

### DIFF
--- a/makedocs
+++ b/makedocs
@@ -14,6 +14,7 @@
 # In order to enable building in newer Qt versions, changes in both the qdoc
 # config files as well as the code comments will be required.
 
+echo '####### $0 starting... #######'
 switched_dirs="false"
 if gitroot=$(git rev-parse --show-toplevel 2>/dev/null); then
 	echo found a git root at $gitroot, running from there...
@@ -59,5 +60,6 @@ echo Generated Docs at "$OUT"
 
 [ $switched_dirs = "true" ] && popd 2>/dev/null || cd -
 
+echo '####### $0 done. #######'
 # always exit 0 for CI runs:
 exit 0


b100dian:
This is awesome - but do you plan to merge the comments you're writing within the existing source files *and* the html output together or split them in two - comments in main, generated docs in the `documentation` branch (much like a `pages` one)?

b100dian:
> 
> GitHub action could build the docs from here, and submit a PR or commit to the other repo.
> 

Exactly. We can define a "pages" repo just for this.
I didn't do this lately with github actions, but I did with woodpecker in codeberg, should be similar https://codeberg.org/sailmates/website/src/branch/main/.woodpecker.yml#L68 (the problem was: sources built a static website that gets commited to pages repo)

Olf0:
> As can be seen in the footer, I am copying the Opal project in placing the generated docs under GNU FDL license. I hope this is OK for everybody.

Yes, though the GNU FDL is very old and the Creative Commons licenses are legally much phrased, plus allow for different variants to pick from ("CC BY-SA" is the equivalent of the GNU FDL).

> Good question - I have been pondering the same.

I would suggest keeping it all in one repo (this one), the chosen location `docs/generated` is fine IMHO.

Similarly I strongly suggest not to depend on multiple service providers, hence do everything on the infrastructure GitHub provides: Publish the generated documentation on GH-pages at `github.io`, for which multiple GH-action examples exist, which can be copy&pasted with a little adaption.

b100dian:
> I would suggest keeping it all in one repo (this one), the chosen location `docs/generated` is fine IMHO.

I don't agree - keeping the same repo yes, keeping in the same branch(es) as the code is way too much compared to, say, localization files.
Same repo, "pages" branch or same org, "pages" repo.

Then again, I'm just commenting here, the doers should have precedence;)


> Similarly I strongly suggest not to depend on multiple service providers, hence do everything on the infrastructure GitHub provides: Publish the generated documentation on GH-pages at `github.io`, for which multiple GH-action examples exist, which can be copy&pasted with little adaption.

Agreed. I was just linking that a similar 'push generated stuff to different repo (or branch)' was possible on another infrastructure.
GH-pages feature is what I have in mind too, and that usually is a different repo or unrelated branch (no common ancestors).

b100dian:
> Now I don't know if gihtub can do different "changes need four-eye review"
> settings per branch, but if not we want a seperate repo, as I'm sure no-one
> will want to review the documentation update PRs all the time.

I'm not sure documentation needs to be reviewed. If it is output from the changes in the code comments, then that can be reviewed instead.

Second topic:
Thanks for cleaning this branch - I am trying to find out what `branches: ["./doc/generated"]` and similarly `   branches: ["documentation"]` actually mean, did you already run one of the new workflows?

Third, question:
Are you aware that in this repo/org you can enable pages and specify a branch or repo for that?
I am not sure I get what "they're currently deployed on nephros.github.io" means, why not do it here?
image

.gitignore
```diff
@@ -1,19 +1,2 @@
...
-*.moc
-patchmanager_interface.*
-patchmanager_adaptor.*
-patchmanager-dialog
-
-.build*
-src/icons/z*
-src/icons/ss
-*.list
+doc/qdoc/sailfish-qdoc-template/
```
b100dian, line 1: hmm I dont think all the other lines need to be removed here

b100dian (approved):
LGTM/Just do it - we will can the documentation itself later when published and patch it up

b100dian:
I would just use the Squash and merge option in this case;)

PR #447, opened 2023-07-05
Add Actions for Documentation publishing

nephros:
Add an Action for publishing Documentation

This will compile the Documentation in qdoc format from `master`, and commit the results to `gh-pages`.
Pages can then be set up to serve that branch.

(`gh-pages` should be created as an empty/orphan branch before using this).

An alternative approach is to instead of pushing to a branch like gh-pages,
push to another branch and use a Pages-Publishing Action to publish.

This approach can be seen e.g. in https://github.com/nephros/patchmanager/blob/56715d0871b9529b998110713fc66b96c941082b/.github/workflows/gendoc-deploy-html-to-pages.yml

~~Depends-On: #442~~
Depends-On: #448


Olf0 (approved):
This is looking good, when reviewing it quickly.

Though I strongly suggest to check, if Docker really needs to be run in privileged mode for this ([line 31](https://github.com/sailfishos-patches/patchmanager/pull/447/files#diff-369c2bf40f83ba663e815184c355b7ba4f2ab34f4c0a9693c12e9891feb1ce73R31)), because that makes local caching impossible (I still hope to get back to this topic sometime).

Additionally I suggest to call the outer Bash (the one called by Docker) also with the option `x`, because that makes debugging easier.

As a "live" example of both, please see https://github.com/sailfishos-patches/patchmanager/blob/master/.github/workflows/ci-on-tags.yml#L39

Olf0:
No, I cannot do that at either!  But not because I do not want to add `endbug/add-and-commit@v*,` there, but because GitHub settings are too sophisticated at places (e.g., here).  Initially I thought this setting is buggy, because it is impossible to alter any of the sub-settings, only the topmost radio-button(s) can be switched to *"Disable actions"* and *"Allow sailfishos-patches actions and reusable workflows"*, but not to *"Allow all actions and reusable workflows"*!  After thinking "I must have set these sub-settings somehow", I comprehended: If an organisation exists, this setting is inherited from there, and on project level one is only allowed to restrict it further.  This all makes sense, but is complicated.  Hence I added `endbug/add-and-commit@v*,` at https://github.com/organizations/sailfishos-patches/settings/actions

Actually there was a reason, why I set this on organisation level: to prevent arbitrary actions to be employed, because they can do everything with a repository.  Hence I think I have to inquire: Have you checked thoroughly that the user / organisation `endbug` is trustworthy?  Plus, have you reviewed this action `add-and-commit`, so you can assume it is technically sound?

Olf0:
Because I am curious, I took the time to look at all three of these actions, even though I knew it would get late.

1. All three actions have a large set of contributors (the one by GitHub the least, but its primary contributors are paid), are well documented, well maintained etc.  After looking at these, I am not afraid any of them will become unmaintained anytime soon, not follow infrastructure changes, is driven by "weirdos" etc.
2. By looking closer now, I finally understood that both [`endbug`'s `Add & Commit`](https://github.com/marketplace/actions/add-commit) and the original [`Git Auto Commit` by Stefan Zweifel](https://github.com/marketplace/actions/git-auto-commit) are quite mighty and generic.
3. Ultimately I personally like the approach via [`Deploy GitHub Pages site` by GitHub](https://github.com/marketplace/actions/deploy-github-pages-site) best, because this action is specifically build for this purpose and technically simpler.  Although it involves one more action to use, they are all provided and maintained by GitHub.  IIRC (when I first researched the topic actions in general, about a year ago) GitHub makes a promise to maintain their released actions for a while and to follow a proper and long deprecation process if they ever decide to stop maintaining one; I believe that is the reason, why they mark the action in question currently still as "[⚠️ Official support for building Pages with Actions is in public beta at the moment.](https://github.com/actions/deploy-pages#scope)", so they can still back out easily, but as it is MIT licensed (as all of GitHub's own actions, IIRC), I would assume it would continue to live even in that case.  So lastly this is just a matter of taste which ones to use and which approach to take, I think.

Olf0:
> * You decided to use `ubuntu-latest`.  This may bring surprises,

Now I remember the whole consideration: As GitHub advises to reference actions including their major release (i.e., as `action@vX`), because this is supposed to ensure API compatibility, and actions usually undergo major release changes much more often than Ubuntu LTS releases do, one has to manually increase these version numbers anyway and then check that everything is still working as intended.  Hence doing this for the Ubuntu LTS release version at least every four years is ensures that one also updates and checks the release versions of the actions used in this cadence.

PR #454, opened 2023-08-16
Slightly overhaul CI workflows

Olf0:
Trigger / Rationale: Trying to download three multi-GB, layered docker images in a single job may result in `docker: write /var/lib/docker/tmp/GetImageBlobXYZ: no space left on device.`, see https://github.com/sailfishos-patches/patchmanager/actions/runs/5871696093/job/15921691170#step:7:45
Hence splitting the "big" `CI on tags` workflow into two jobs and slightly enhancing both CI workflows.

### DIFF
--- a/.github/workflows/ci-on-pull_req.yml
+++ b/.github/workflows/ci-on-pull_req.yml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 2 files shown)


Olf0 (commented):
I hope it is O.K. so far.

PR #455, opened 2023-08-16
[gendoc-create-html_qt56.yml] Omit check to "fix" explicitly empty input

Olf0:
Reference: https://github.com/sailfishos-patches/patchmanager/pull/449#issuecomment-1681156014


Olf0:
I just ran it, because I missed to do that before: https://github.com/sailfishos-patches/patchmanager/actions/runs/5930615413

Is fine, merging.

PR #459, opened 2024-02-25
[UI] Use icons instead of Switches

nephros:
This may prove controversial, but replace the glass switches with icon buttons

Sorry oldschool ui lovers (and I am one myself).

Much cleaner look of the patch list. Also, now all Patches get an icon, not
only those shipping one with the patch.

- [UI] Lock pulley menu while dragging
- [UI] Remove glass switch, use icons
- [UI] Use Icons instead of Switch in Restart Dialog

### DIFF
--- a/src/qml/PatchManagerPage.qml
+++ b/src/qml/PatchManagerPage.qml
@@ -528,7 +528,9 @@ Page {
                     radius: 0.14
                     falloffRadius: 0.13
                     visible: (down || busy || patchObject.details.patched)
-                    color: (down || busy || patchObject.details.patched) ? Theme.primaryColor : Theme.rgba(Theme.secondaryColor, Theme.opacityLow)
+                    color: (down || busy || patchObject.details.patched)
+                        ? Theme.rgba(Theme.primaryColor, Theme.opacityLow)
+                        : Theme.rgba(Theme.secondaryColor, Theme.opacityLow)
                     Behavior on color { FadeAnimation {} }
                 }
 
@@ -567,7 +569,6 @@ Page {
                     icon.sourceSize.width: Theme.iconSizeSmallPlus
                     icon.height: Theme.iconSizeSmallPlus
                     icon.width: Theme.iconSizeSmallPlus
-                    icon.opacity: highlighted ? 1.0 : Theme.opacityLow
 
                     palette.primaryColor: Theme.secondaryColor
                     palette.highlightColor: Theme.primaryColor


b100dian:
Looks better than today indeed!

The only worry is if enabled/disabled is signalled by changing the icons color luminance, since I have to test it, my intuiltion says there a thinner and thicker paths in those icons that may look lighter as in disabled. I assume we don't have an outline or glow effect at hand to "offset" this?

Olf0:
> There is https://doc.qt.io/qt-5/qml-qtgraphicaleffects-glow.html
> 
> but is it worth it?

Oh, a cool, mighty, but costly effect.  Would have to use the `cache` property (as these icons are small and not animated) and might also need alter the glow colour depending on the background colour (i.e. dark vs. light ambiences: a white glow is almost invisible on a bright background).

Is that easy to use / implement?

b100dian:
Thanks for the code sample. This is so fun to play with:) Here's my try:

```
diff --git a/src/qml/PatchManagerPage.qml b/src/qml/PatchManagerPage.qml
index e7a7bed..a9dc386 100644
--- a/src/qml/PatchManagerPage.qml
+++ b/src/qml/PatchManagerPage.qml
@@ -38,6 +38,7 @@
 import QtQuick 2.0
 import Sailfish.Silica 1.0
 import Nemo.Configuration 1.0
+import QtGraphicalEffects 1.0
 import org.SfietKonstantin.patchmanager 2.0
 
 /*! \qmltype PatchManagerPage
@@ -597,6 +598,16 @@ Page {
                     Behavior on icon.opacity { PropertyAnimation {
                         duration: 1200; alwaysRunToEnd : true; easing.type: Easing.OutBack
                     }}
+                    Glow {
+                        visible: patchObject.details.patched
+                        cached: true
+                        anchors.fill: parent.icon
+                        radius: 20
+                        samples: 20
+                        color: Theme.highlightColor
+                        source: parent.icon
+                       opacity: 0.5
+                    }
 
                 }
 
```

b100dian:
However I must admit I agree with the fact that the effect may be suboptimal, that I tested the original version before and it was pretty usable, and that the opacity thing is indeed an advantage (though that remains true if you want to animate the glow too).

So while this was fun I think (with my first comment about duplicated homescreen entry) we should just go with your version as it is cleaner and glow just adds more un-sailfish-y things and no-one here is a designer;P

b100dian (approved):
:sparkles:

Olf0:
> Here's my try:

Oh, both screenshots look fantastic.  I did not expect the glow effect to be so nice on both, a bright and dark background, simply by choosing `color: Theme.highlightColor` for the glow effect.

Besides the pretty look, I do appreciate that the visual contrast between on and off state is well increased, despite it technically still is `opacity: 0.5` (but plus glow effect here).

I also prefer the lack of any cyclic animations, any continuous pulsing or blinking (even softly) gets on my nerves very easily; I know that I am much more susceptible for such stimuli, they quickly irritate and annoy me.

But as you both prefer the rather simple look, this is also fine for me, if only the visual contrast between on and off state would be larger: Maybe `opacity: 0.4` (down from `opacity: 0.5`, AFAICS) for icon and the following line of text archives that.
<sub>BTW, IMO the simple look is only more "sailfishy" relative to SFOS 1 & 2, but not really SFOS ≥ 3.</sub>

b100dian:
Hmm so this moves the dots to the left? That is also the reason there is overlap with page navigation dot? Hmm not entirely sure how this is better..

b100dian:
I see, so if it is like email I assume you have scrolled before screenshotting, tbat was nit clear, why the overlap :+1: . Again, we can get this in and see if there is feedback and we have several alternatives to fast follow up if so.

PR #460, opened 2024-05-05
[QML] Align WebPatch Details design with Patch Details

nephros:
This makes the details for a patch from the catalog appear similar to that of an installed patch


b100dian (approved):
LGTM

(Together with #459 you could make a release y'know)

PR #461, opened 2024-05-06
[QML] Make initial loading screen less stuttery

nephros:
There's lots going on on the initial page when launched the first time.
This lead to all kinds of ugliness like stuttering and objects overlaying each other.

This tries to improve things, or at least hide it from view ;)

- Make ViewPlaceholder only appear if there are indeed no patches
- Use different Transitions
- Remove Add Transition

### DIFF
--- a/src/qml/PatchManagerPage.qml
+++ b/src/qml/PatchManagerPage.qml
@@ -96,23 +96,28 @@ Page {
         interval: 1500
         repeat: false
         running: true
-        onTriggered: console.debug()
-        onRunningChanged: console.debug(running)
+        onTriggered: {
+            if (PatchManager.updatesNames.length > 0) {
+                hintTimer.start()
+            }
+        }
+    }
+
+    Timer {
+        id : hintTimer
+        interval: 3000
+        repeat: false
+        running: false
     }
 
     Connections {
         target: PatchManager
         onUpdatesChanged: {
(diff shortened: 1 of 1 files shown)


src/qml/PatchManagerPage.qml
```diff
@@ -123,88 +128,17 @@ Page {
...
-        }
-
-        pulleyAnimation.start()
-    }
-
-    /*! \qmlproperty real PatchManagerPage::pullDownDistance
-        \internal
-        Amount of space the pulley "pops down" when it's showing a hint, e.g the result of showUpdates().
-
-        \warning This is probably broken in recent SFOS versions (?).
```
b100dian, line 147: I actually very recently saw this when you updated the Bluetooth patch, and thought: wow, that's nice and wait, that's a bit off -> probably the warning above.

Is this completely removed now? I am not able to see it replaced by anything.

b100dian: Ah, you mentioned it is broken/not anymore present, it is just that the comment wasn't attached to the diff. What can I say, I just _saw_ it working today :))

Olf0:
> And: I usually start PM from a topmenu link not by going through Settings. There the loading behaviour is slightly different, and the hint indeed is broken.

IMO both ways are legitimate and hence should work, right?

PR #462, opened 2024-05-06
[UI] Add option to show only updates

nephros:
Adds a Setting option (client-only) which when enabled will hide all
non-updatable patches from the web catalog list, if there are any updates.

- hide all non-updatables while updates available
- Emulate a ListItem


Olf0 (approved):
LGTM, but looking at it is not really reviewing code.  Nevertheless, approved to move forward.

PR #464, opened 2024-05-28
[SFOS 4.6] Support new theme icon location

nephros:
This took me way longer to figure out than should be necessary.

- Just use SAILFISH_SVG2PNG for theme icons
- Package any theme location
- Hardlink the icon source, add README
- Add svg to installs
- Fix packaging

Background: Sailfish OS 4.6 changes the default path for "theme" icons, such as
the ones used for Settings plugins (which we are).

Additionally, the layout of the source dir for svgs to convert has changed or
the sailfish-svg2png tool.

This drops amost all of src/icons/icons.pro, as that was a verbatim copy of the
upstream sailfish-svg2png .prf file, and just uses that prf file as intended.

I haven't checked, but this *may* mean we lose some of the icon scales for older
packages, but it should't realy make a noteworthy difference.

### DIFF
--- a/src/icons/README.md
+++ b/src/icons/README.md
@@ -8,5 +8,5 @@ For all SailfishOS versions up to v4.6 (and possible also Meego):
 For SailishOS 4.6 and higher
 
  - svg files must be in svgs/icons and svgs/icons-monochrome
- - png files will be in /usr/share/sailfish-default/silica/zX.Y/icons
+ - png files will be in /usr/share/sailfish-default/silica/zX.Y/icons and /usr/share/sailfish-default/silica/zX.Y/icons-monochrome
 


Olf0:
Thanks for the fixup.  I really start to hate them for doing such breaking changes as altering arbitrary names regularly.

I will try to take a look at it and comprehend what you researched and implemented (I may need this for other projects, too) next weekend.

rpm/patchmanager.spec
```diff
@@ -291,7 +290,13 @@ exit 0
 %{_datadir}/jolla-settings/entries/%{name}.json
 %{_datadir}/%{name}/icons/icon-m-patchmanager.png
 
-%{_datadir}/themes/%{theme}/meegotouch/*/icons/*.png
+# pre 4.6:
+# /usr/share/themes/sailfish-default/meegotouch/zX.Y/icons/*.png
+# 4.6 and higher:
+# /usr/share/themes/sailfish-default/silica/zX.Y/icons/*.png
+# /usr/share/themes/sailfish-default/silica/zX.Y/icons-monochrome/*.png
+%{_datadir}/themes/sailfish-default/*/*/*/*.png
```
b100dian, line 298: based on the above, shouldn't this be ...`/*/*/icons/*.png`?

b100dian:
Thanks for the quick fix!

b100dian (approved):

Olf0 (commented):
LGTM

PR #452, opened 2023-08-15
[patchmanager.cpp] Reintroduce two dummy methods for documentation …

Olf0:
… purposes only in order to silence `qdoc` warnings WRT undocumented classes, methods and functions.
Reference: The core aspect of the whole back and forth in PR #451: https://github.com/sailfishos-patches/patchmanager/pull/451

PR #465, opened 2024-05-28
[Doc] Remove/Improve build-breaking "Dummy" method docs

nephros:
Turns out these methods are not so dummy after all. (See #452)

One is a Signal, which shouldn't have an implementation at all.

The other is indeed dead code, but lack of usage o its parameter broke the
build. Instad of wrapping it in `Q_UNUSED`, lets just not have an
implementation, just the qdoc comment.

This now does silence qdoc, and does not break the build.

### DIFF
--- a/src/qml/patchmanager.cpp
+++ b/src/qml/patchmanager.cpp
@@ -614,7 +614,10 @@ void PatchManager::doVote(const QString &patch, int action)
     putSettingsSync(QStringLiteral("votes/%1").arg(patch), action);
 }
 
-/*! \internal lets not spoil the fun (or the eggs!). */
+/*!
+   \internal
+   lets not spoil the fun (or the eggs!).
+*/
 void PatchManager::checkEaster()
 {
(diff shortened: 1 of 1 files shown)


Olf0:
Will look at this, try to comprehend the storyline which developed over years and approve, soon™.

Olf0 (approved):
IMO this confirms to stay away from any "easter shit" in the first place: It is additional, superfluous code, which causes a maintenance burden. :stuck_out_tongue:
Well, never mind my ranting.

Seriously:
I do not fully comprehend why you have taken out the textual pointers to the function declarations in the corresponding header files and to the locations these functions are used, which were intended not to loose track of these places which will have to be tackled when fully eliminating this unnecessary code.  Knowing GitHub-search better now, I know they are easy to find again, but still someone will have to perform this (re)search again, then.

Still I approve this PR as I do not have an overly strong opinion about this, but I would be glad if you reconsider keeping the detailed comments in or explicitly name a reason why you believe they must go.

PR #471, opened 2024-11-25
[patchmanager.spec] Specify icon paths more strictly

Olf0:


Olf0:
The vast degree(s) of freedom for the icon paths was frightening me; well I know, technically it is only about what is copied from RPM's `buildroot`, but `%{_datadir}/themes/sailfish-default/*/*/*/*.png` … frightens me a bit.

A more rationale point is, that `*/*/*/*.png` is hard for a reader to retrace, which files are meant to become deployed here.

BTW, I restricted my self to specify `{_datadir}/themes/sailfish-default/*/z*/icons*/*.png` while I was considering `{_datadir}/themes/sailfish-default/{silica,meegotouch}/z?.?*/icons*/*.png`.  If a reviewer prefers the latter, I will happily use that; I was just afraid this would be regarded as "olf is nitpicking again" (and I never used brace expansion in a spec file's file list before).

Olf0:
> LGTM.
> 
> > BTW, I restricted my self to specify `%{_datadir}/themes/sailfish-default/*/z*/icons*/*.png` while I was considering `%{_datadir}/themes/sailfish-default/{silica,meegotouch}/z?.?*/icons*/*.png`. If a reviewer prefers the latter, I will happily use that; I was just afraid this would be regarded as "olf is nitpicking again" (and I never used brace expansion in a spec file's file list before).
> 
> I'm not sure the path expansion with curly braces would work with RPM.

No, it does not (support brace expansion), while the wildcards `*` and `?` are fully supported: https://rpm-software-management.github.io/rpm/manual/spec.html#shell-globbing

P.S.: I am way behind many of my schedules (privately, at work and here), I will try to catch up after Christmas. Sorry!

Issue #468, opened 2024-11-25
[Bug] 64bit succeeds but applies 32bit patch


ichthyosaurus:
This bug report is just a placeholder to track a bug discussed in the forum: https://forum.sailfishos.org/t/patches-by-ichthyosaurus/15387/119

The [color emoji patch](https://openrepos.net/content/ichthyosaurus/patch-stock-emoji-keyboard-colors) fails to apply on 64bit devices running SFOS >= 4.6 because 64bit mangling succeeds but PM still applies the 32bit code.

ichthyosaurus:
Wow, thank you for this effort! I don't have a 64bit device though, so I can only rely on others for testing. Is it ok if I post your rpm in the forum? (Or just release it if it works on your end...)

(Just a thing that I noticed while reading your posts: you mentioned dangling symlinks, and `QFileInfo::exists` seems to be used heavily. That function returns false for broken symlinks - that could cause bugs unrelated to this one.)

ichthyosaurus:
> Testing on 32bit is fine too, to make sure it doesn't break anything.

Will do!

> I'd politely ask to not put the rpm or link on the forum right now. It can cause problems with updating to a proper release version to use rsndom OBS versions.

I figured as much, that's why I asked :)

ichthyosaurus:
Everything seems to work fine, except that Patchmanager lost its icon in the settings app. Tested on Xperia X (32bit), SFOS 4.5.0.24.

ichthyosaurus:
> Erm, that would be bad indeed. I can't think of a thing why that could happen though...

yes it's weird and I don't see how that could be related...

PR #472, opened 2024-11-25
[ci-on-tags.yml] Redefine LATEST sensibly

Olf0:
For details see new comment in the altered file.

### DIFF
--- a/.github/workflows/ci-on-tags.yml
+++ b/.github/workflows/ci-on-tags.yml
@@ -11,8 +11,9 @@ on:
   # Rather set a new tag in the format N/X.Y.Z (e.g., 1/0.6.3) to build a release originally tagged with 0.6.3 again.
 
 env:
-  # For the latest available docker image, see https://github.com/CODeRUS/docker-sailfishos-platform-sdk
-  LATEST: 4.5.0.16
+  # For the available docker images, see https://github.com/CODeRUS/docker-sailfishos-platform-sdk
+  # But LATEST means here, known to run on the latest SailfishOS release: Binaries for 4.3.0 are know to run on 5.0.0
+  LATEST: 4.3.0.12
   # Do not wait up to the default of 10 minutes for network timeouts in a workflow which runs at most 20 minutes.
   SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
 
@@ -174,4 +175,3 @@ jobs:
 #branding:  # "Invalid workflow file: Unexpected value 'branding'"; maybe action MUST be located in '/' or the name MUST be action.yml or both, see e.g., https://github.com/actions/cache/blob/main/action.yml#L37
 #  icon: 'gift'
 #  color: 'purple'
-
--- a/.github/workflows/ci-on-tags.yml
+++ b/.github/workflows/ci-on-tags.yml
@@ -12,7 +12,7 @@ on:
 
 env:
   # For the available docker images, see https://github.com/CODeRUS/docker-sailfishos-platform-sdk
-  # But LATEST means here, "known to run on the latest SailfishOS release": Binaries for 4.3.0 are known to run on 5.0.0
+  # But LATEST means here, "known to run on the latest SailfishOS release": Binaries for 4.3.0 are known to run on 5.0.0.
   LATEST: 4.3.0.12
   # Do not wait up to the default of 10 minutes for network timeouts in a workflow which runs at most 20 minutes.
   SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
--- a/.github/workflows/ci-on-tags.yml
+++ b/.github/workflows/ci-on-tags.yml
@@ -12,7 +12,7 @@ on:
 
 env:
   # For the available docker images, see https://github.com/CODeRUS/docker-sailfishos-platform-sdk
-  # But LATEST means here, known to run on the latest SailfishOS release: Binaries for 4.3.0 are know to run on 5.0.0
+  # But LATEST means here, "known to run on the latest SailfishOS release": Binaries for 4.3.0 are known to run on 5.0.0
   LATEST: 4.3.0.12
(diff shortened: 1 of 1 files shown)


Olf0:
BTW / OT:
> Which means as long as there is no docker env for 4.6 and higher, icons will end up on the "old" place, and if the resulting packages are installed, there will be no icon for Patchmanager in Settings, and the Top menu.
> 
> I have worked around this issue on OBS by "backporting" (i.e. adding) sailfish-svg2png-0.4.1 from 4.6/5.0 to the build env: https://build.sailfishos.org/project/show/home:nephros:devel:patchmanager. Builds in that OBS project will use it, and therefore package the icon files correctly got 4.6 and higher.

What does that mean for Storeman build at the SFOS-OBS without such workarounds (or FileCase / Flowplayer)?  Do they also show no icons (or only no in-app icons?) without this workaround?  Shouldn't this workaround then be better available at SailfishOS:Chum(:Testing), too?
<sub>Did I ever mention that I hate Jolla for their permanent stream of app-breaking changes out of carelessness / negligence? :wink:</sub>

Olf0:
> I haven't checked these apps, but all of this only applies to "theme" icons under `/usr/share/themes`.

Thank you for the clarification!  Yes, now I see this is nothing I have to worry about.



> Now, to solve the whole current icon mess, we _could_, in Patchmanager, move the icon location to `/usr/share/icons`. I believe that will work also for our Settings plugin. This way we would be compatible with all releases. 

:+1:
That sounds like a cool way out, in order to avoid another `sailfish-version` dependent switch.

> At least until someone decides to change the icon theme path away from `hicolor`. I think that's unlikely though, as that name is probably defined by the FDO desktop standard.

Yes, it is by [FDO's icon theme specification]](https://specifications.freedesktop.org/icon-theme-spec/latest/#directory_layout).  SO that looks like a sound approach.

Olf0:
Back to the original topic, i.e. the one of this PR:
> All in all, I do not exactly understand what the requirements for the CI workflows are, …

Practically I might add a true `LATEST` in another PR and use the string `latest-SDK` in the generated artefacts.

Olf0:
> > At least until someone decides to change the icon theme path away from `hicolor`. I think that's unlikely though, as that name is probably defined by the FDO desktop standard.
> 
> Yes, it is by [FDO's icon theme specification](https://specifications.freedesktop.org/icon-theme-spec/latest/#directory_layout): `$XDG_DATA_DIRS` resolves to `/user/share` (all elements on OpenSUSE LEAP: `~/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share:/usr/share`) and a `hicolor` directory there is required, so `/usr/share/icons/hicolor` must exist. Hence that appears to be a sound approach.

See also the corresponding remark by @ichthyosaurus in https://github.com/sailfishos-patches/patchmanager/issues/468#issuecomment-2506494808.

Olf0:
Unfortunate, but "it is what it is".

> I realise that I do not use sailfish-svg2png of which versions ≥ 0.4.1 likely deploy the generated PNGs to the new paths.

Is this understanding I gathered correct (please confirm)?
If so, IMO it makes sense that I:
* Build (also) with `SDK-4.6` in the `ci-on-tags.yml` workflow.
* Use a `sailfish-version` based switch in the `%files` section of the spec file (I am starting to become an expert on this, unfortunately) to reference the correct theme icon directories.

Does this plan sound reasonable for you?

Olf0:
> > Unfortunate, but "it is what it is".
> > > I realise that I do not use sailfish-svg2png of which versions ≥ 0.4.1 likely deploy the generated PNGs to the new paths.
> > 
> > 
> > Is this understanding I gathered correct (please confirm)?
> 
> Versions of sailfish-svg2png >0.4.1, which are used in qmake .pro files (via `CONFIG += sailfish-svg2png`), will deploy to the new paths, older to the old paths. Path meaning `/usr/share/themes/foo`.
> 
> One can also use `sailfish-svg2png` in other ways, by calling the program manually instead of relying on the shipped qmake configuration. (This can be used to populate `/usr/share/icons/hicolor/NxN` with correctly scaled icons generated from a SVG file.) By forgoing the shipped configuration, one can put the results anywhere.
> 
> At Patchmanager, before #464, we used a _part_ of the shipped configuration, but still called the binary manually. This was a bit superfluous, as we basically did just the same thing as the shipped configuration does. Which is why #464 reduces the .pro file to just a few lines. (But that may have been different in older releases.)
> 
> > If so, IMO it makes sense that I:
> > ```
> > * Build (also) with `SDK-4.6` in the `ci-on-tags.yml` workflow.
> > ```
> 
> [EDIT:] most of the following proved to be irrelevant. See the next comment.
> 
> I guess.
> 
> Whichever way we deal with this, it all boils down to the fact that we have to specify the version the rpm is _going to be installed on_ during the build process. If install-os-version === build-host-version, the current setup works for all releases.
> 
> There's a couple of things we could do differently, e.g. doing the generation of the icons manually independently of build-host-version, copy them around after generation etc., but with all those variants we always have to specify the install-os-version in some way.
> 
> The only point where we can be sure what to do is at RPM install time, so during the `%post` and related phases. There we could copy the files to the correct location. But that's ugly, and it makes the filesystem contents not match with the RPM contents.
> 
> We could even do something more ugly, and `Require: sailfish-svg2png`, and _generate_ the PNGs at install-time :D, and use `%ghost` entries in the RPM `%files` section.
> 
> > ```
> > * Use a `sailfish-version` based switch in the `%files` section of the spec file (I am starting to become an expert on this, unfortunately) to reference the correct theme icon directories.
> > ```
> 
> I'm not sure this is required, if the right many-asterisked paths are used. Currently I think we are good here.

Ack: Sorry, I already forgot how I designed PR #471, because the last weeks have been extremely busy for my.

So with that being understood as superfluous, the only thing left is also building with the SailfishOS-SDK for 4.6.0.

PR #474, opened 2024-11-26
[Log] Reduce noise, improve readability

nephros:
This attempts to make reading the journal easier.

Drop some noisy messages down to Debug level, make some other messages
more readable by omitting Q_FUNC_INFO.

- **Logging: Drop function signature from user-visible messages**
- **Logging: decorate DBus warnings**
- **Logging: Downgrade some noisy messages to Debug level**


Olf0 (approved):
LGTM

PR #475, opened 2024-11-26
Test Cases: Add a git-style file-add patch

nephros:
Test to check whether git-style diffs behave correctly.

Related: #468


Olf0:
Oh, likely another case of "do not use GitHub's "review" functionality, because it creates messages invisible for some!

My review was / is:

**[`tests/patching/add-a-file-git-style/makediff.sh`](https://github.com/sailfishos-patches/patchmanager/pull/475/files/8b6dffd7bebcc72e65c7c3e8e7fa113092cc1e80#diff-1b241eb20d3819a76158cd884daa87cd891a40d99dd6e346aa598d17f9546c26)**
> @@ -0,0 +1,17 @@
> #!/bin/sh

@Olf0:
IIRC pushd, popd and dirs are bashisms.
Edit: I fail to find them at https://pubs.opengroup.org/onlinepubs/9699919799/utilities/contents.html

busybox-ash in SailfishOS may be configured to support them when Jolla compiles it, have you checked that?

Still I suggest to call #!/bin/bash as interpreter to make things obvious (busybox-ash in SailfishOS is accompanied by a bash-symlink package), if Jolla's busybox-ash configuration supports them.

PR #476, opened 2024-11-27
Improve 64bit mangling, support file creation

nephros:
Fixes: #468

For details, see some of the comments in the file, and the analysis in issue #468, especially
https://github.com/sailfishos-patches/patchmanager/issues/468#issuecomment-2502156142

- **Add a dedicated method for path mangling**
- **Replace the inline mangling with the new method.**
- **Also mangle paths for newly created files.**

### DIFF
--- a/src/bin/patchmanager-daemon/dbus/org.SfietKonstantin.patchmanager.xml
+++ b/src/bin/patchmanager-daemon/dbus/org.SfietKonstantin.patchmanager.xml
@@ -107,6 +107,12 @@
         <method name="getLoaded">
             <arg name="loaded" type="b" direction="out" />
         </method>
+        <method name="restorePatchList">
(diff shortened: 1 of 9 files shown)


Olf0 (commented):
This PR in its final form looks absolutely reasonable to me, but (as often stated) I am not a C++ coder.

The corresponding analysis and code development was actually fun to read after spending an exhausting working day mostly in teleconferences:
> the analysis in issue https://github.com/sailfishos-patches/patchmanager/issues/468, especially https://github.com/sailfishos-patches/patchmanager/issues/468#issuecomment-2502156142

Still I would prefer @b100dian to review this, especially as he was heavily involved in creating the 32-/64-bit Patch-mangling in the first place, IIRC.

If @b100dian has no time for a review, I sure will approve this, just ping me (either @b100dian or @nephros).

PR #479, opened 2024-12-03
Package icons as subpackages

nephros:
Supporting two different filesystem locations depending on Sailfish OS version
while keeping the number of build environments to a minimum is problematic.

Instead of deciding where to install and what to package at build time, package and ship
both variants and let the package manager decide which is the right one.

See also: https://github.com/sailfishos-patches/patchmanager/pull/472#issuecomment-2499936641

- **Package icons as subpackages**
- **Copy generated icons around**
- **Add Obsoletes, in order to provide an upgrade path**

Things to test:

 - [x] picks the correct package on SFOS 4.6 and/or 5.0
 - [ ] picks the correct package on SFOS 4.5 and lower
 - [ ] picks the correct package when updating from an installed Patchmanager version to the one including this change 
 - [ ] Shows icon correctly after installation (for all of the above)
 - [ ] Removes icons at the `meegotouch` path and installs them at the `silica` location after/during an update from a lower version to SFOS 4.6
   SFOS GUI upgrades usually remove Patchmanager anyway AFAIK, so this last test may not be that important.

### DIFF
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -59,6 +59,7 @@ BuildRequires:  pkgconfig(popt)
 Summary:    Icon files for %{name}
 BuildArch:  noarch
 Requires:   sailfish-version >= 4.6.0
+Requires:   %{name} >= %{version}
 Provides:   %{name}-icons
 Conflicts:  %{name}-icons-meegotouch
 Obsoletes:  %{name}-icons-meegotouch
@@ -67,6 +68,7 @@ Obsoletes:  %{name}-icons-meegotouch
 Summary:    Icon files for %{name}
 BuildArch:  noarch
 Requires:   sailfish-version < 4.6.0
+Requires:   %{name} >= %{version}
 Provides:   %{name}-icons
 Conflicts:  %{name}-icons-silica
 


Olf0:
From [issue #472 comment-2514204204](https://github.com/sailfishos-patches/patchmanager/pull/472#issuecomment-2514204204):
> That should shift the burden to the package management solver and thereby solve (heh) our problems with build/install/SDK/release versions.

Cool, proves (agaib) that ["we can solve any problem by introducing an extra level of indirection"](https://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering)™.

**But**, as [I have understood again](https://github.com/sailfishos-patches/patchmanager/pull/472#issuecomment-2521871180) that just also building with the SailfishOS-SDK for 4.6.0  in the `ci_on_git-tags` workflow completely resolves this issue, **and** exactly that already happens at SailfishOS:Chum (i.e. the SailfishOS-OBS), I think this is over-engineered.
I.e. I like it technically, but it introduces additional complexity which makes maintenance harder (especially if other people should continue maintaining Patchmanager, as it already happened twice).

Hence currently I do not think this is needed.  But maybe I have not understood how this structural change might be helpful in the future.
I.e. if this PR really solely resolves the minor issue of the `ci_on_git-tags` workflow currently not generating RPM packages suitable for SailfishOS ≥ 4.6.0, I strongly prefer to resolve this issue with a minor adaptation of [the `ci_on_git-tags` workflow configuration file](https://github.com/sailfishos-patches/patchmanager/blob/master/.github/workflows/ci-on-tags.yml).

Olf0 (commented):
From [issue #472 comment-2514204204](https://github.com/sailfishos-patches/patchmanager/pull/472#issuecomment-2514204204):
> That should shift the burden to the package management solver and thereby solve (heh) our problems with build/install/SDK/release versions.

Cool, proves (again) that ["we can solve any problem by introducing an extra level of indirection"](https://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering)™.

**But**, as [I have understood again](https://github.com/sailfishos-patches/patchmanager/pull/472#issuecomment-2521871180) that just also building with the SailfishOS-SDK for 4.6.0  in the `ci_on_git-tags` workflow completely resolves this issue, **and** exactly that already happens at SailfishOS:Chum (i.e. the SailfishOS-OBS), I think this is over-engineered.
I.e. I like it technically, but it introduces additional complexity which makes maintenance harder (especially if other people should continue maintaining Patchmanager, as it already happened twice).

Hence currently I do not think this is needed.  But maybe I have not understood how this structural change might be helpful in the future.
I.e. if this PR solely resolves the minor issue of the `ci_on_git-tags` workflow currently not generating RPM packages suitable for SailfishOS ≥ 4.6.0, I strongly prefer to resolve this issue with a minor adaptation of [the `ci_on_git-tags` workflow configuration file](https://github.com/sailfishos-patches/patchmanager/blob/master/.github/workflows/ci-on-tags.yml) (I will do that soon™).

Olf0:
> I strongly prefer to resolve this issue with a minor adaptation of [the `ci_on_git-tags` workflow configuration file](https://github.com/sailfishos-patches/patchmanager/blob/master/.github/workflows/ci-on-tags.yml) (I will do that soon™).

Done per [PR *"[ci-on-tags.yml] Also build with SDK for 4.6.0"* #485](https://github.com/sailfishos-patches/patchmanager/pull/485/files).

PR #484, opened 2025-01-12
[ci-on-pull_req.yml] Omit misleading "s" in comment

Olf0:
… and up `actions/cache` to `@v4`, see https://github.com/actions/cache#readme.

### DIFF
--- a/.github/workflows/ci-on-pull_req.yml
+++ b/.github/workflows/ci-on-pull_req.yml
@@ -29,7 +29,7 @@ jobs:
   build:
     runs-on: ubuntu-24.04
     env:
-      # Solely builds for i486 on 3.4.0, because of https://github.com/sailfishos-patches/patchmanager/pull/437#issuecomment-1615317003
+      # Solely build for i486 on 3.4.0, because of https://github.com/sailfishos-patches/patchmanager/pull/437#issuecomment-1615317003
       ARCH: i486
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-on-pull_req.yml
```diff
@@ -29,7 +29,7 @@ jobs:
   build:
     runs-on: ubuntu-24.04
     env:
-      # Solely builds for i486 on 3.4.0, because of https://github.com/sailfishos-patches/patchmanager/pull/437#issuecomment-1615317003
```
b100dian, line 32: Was it a noun plural or a verb :P ?

> [It] solely builds ..

Or..

> Built solely..

Olf0: **a.** A verb

**b.**
> [It] solely builds ..
… is exactly what was wrong before: It sure also builds for newer SailfishOS releases.

> Built solely..
… is almost correct, but I hesitate to use the past tense, because this is the build configuration, which determines how Patchmanager **is** build: In the past, now and in the future.

But I gladly follow the idea to swap the "build" and "solely": [commit `9a16ce6`](https://github.com/sailfishos-patches/patchmanager/pull/484/commits/9a16ce6d76f2649dfc3a9f87706cee534451c31c)

b100dian (approved):
Nice discussion ;)

PR #495, opened 2025-03-14
[pm_apply] test for unclean patches and warn about it

nephros:
See code.

There are some patches in the wild which contain unwanted changes.
This will make some of these issues show up in the log.


Olf0 (approved):
LGTM

PR #497, opened 2025-03-31
Ship a debug configuration file for the daemon

nephros:
This is intended to make debugging easier.
It should contain any supported debug environment variables so we don't have to
look them up in the code or documentation

The file is read by `/usr/lib/systemd/system/dbus-org.SfietKonstantin.patchmanager.service` (so, the main daemon)
to set its environment.


Olf0 (approved):
LGTM

Issue #480, opened 2024-12-15
[Bug] Dialog should wait for PIN entry window(s)


nephros:
**SailfishOS VERSION**: 4.x, 5.0, likely all
<br />**HARDWARE**: Sony Xperia 10 iii, Jolla C2, likely any
<br />**PATCHMANAGER VERSION**: 3.2.11
 
<br /> 

#### BUG DESCRIPTION

See https://forum.sailfishos.org/t/sim-pin-remaining-attempt-count-may-be-wrong/21409

After initial boot, Patchmanager should probably wait for both the Lock Code as well as any SIM PIN entry dialogs before starting its "apply all active patches" countdown.

Olf0:
As [you asked there](https://forum.sailfishos.org/t/sim-pin-remaining-attempt-count-may-be-wrong/21409/2):
> To fix this on the Patchmanager side: is there a way to detect from an application that the SIM PIN dialog is active currently??

Provided that the SIM PIN dialogue is started by a `systemd` unit, one may query if that unit is currently active.  But it depends on the implementation if such a unit just triggers the dialogue or stays active until the dialogue is closed.
Side note: AFAIU this dialogue may appear at any time, although in practise it is usually displayed before the "Launcher" is started, this is nothing one should rely on AFAIU.  Consequently simply delaying the start of Patchmanager (e.g. until the "Launcher" is fully up) does not resolve this issue reliably.

After all, this issue is clearly a bug in the implementation of the SIM PIN dialogue, as it does not query the number of remaining tries from the SIM card, but uses its own. independent counter.

PR #500, opened 2025-04-19
Dialog: make cancel timeout configurable

nephros:
- Dialog: make timeout configurable
- Dialog: use non-deprecated import
- Dialog: allow any orientation
- Dialog: Fix a warning

Contributes-To: #480

### DIFF
--- a/src/bin/dialog/dialog.qml
+++ b/src/bin/dialog/dialog.qml
@@ -218,7 +218,7 @@ ApplicationWindow {
             }
 
             CoverActionList {
-                enabled: appWindow.remorseItem.pending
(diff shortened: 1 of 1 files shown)


Olf0 (approved):
LGTM

Olf0:
@nephros, when trying to translate [string \#111 at Transifex](https://app.transifex.com/coderus/patchmanager3/translate/#de/translations-settings-patchmanager-ts--master/578036038?q=translated%3Ano), it appears to me that a "when rebooting" is missing [in the original source string](https://github.com/sailfishos-patches/patchmanager/blob/master/src/qml/SettingsPage.qml#L124) to make sense, correct?
Maybe we can even drop the lengthy "for the first time" and replace it with "when rebooting"?

PR #506, opened 2025-05-21
[systemd] only run update timer on normal boot

nephros:
Noticed that this is activated too early, e.g. when coming out of
"actdead" mode.

We really need to run only on a regularly booted system.


Olf0 (commented):
This makes sense, I observed this misbehaviour, too.

But shouldn't we rather wait until some GUI services run?
(I can look up a concise suggestion, if you like.)

Does it make sense to start the Patchmanager Daemon when her is no GUI?

Olf0:
O.K., then on to the technical aspects:
- `After=` is a purely temporal ordering.  I.e. if unit B states `After=A`, it will be started when A has failed to start, and IIRC even when A is unscheduled and inactive (and surely also also when A is up and running).  Effectively it states "After a scheduled decision on A has been met."  Hence it is most often used in combination with `Depends=` on other units, devices etc.
- But for `.target`s usually `PartOf=` is used (that is what `PartOf=` is made for), not the `Depends=A` & `After=A` combination which is used for `.unit`s, `.device`s etc.
  This why I posed PR #515.
- ***But*** maybe that is not the intended behaviour.  While `After=` alone is pretty sure incomplete, I may do what you intended, because:
- The default dependencies cause this unit not to be scheduled early.
- When this unit is scheduled, most likely the `multiuser.target` is already scheduled, too.
- In this situation I believe that `After=multiuser.target` lets this unit wait until all units referencing `PartOf=multiuser.target` have been dealt with.  If that is really needed, I might ponder about how to express this dependency correctly and come up with a suggestion.

Olf0 (approved):
As I assume you tested this PR applied to one of your SailfishOS installations (initially without, and later with my PR #515 applied), I approve this PR by an always implicitly vague ( :wink: ): **LGTM**.

PR #513, opened 2025-07-02
[translations/settings-patchmanager.ts] Update by runnning `lupdate`

Olf0:
Specific command used on OpenSUSE:
`/usr/lib64/qt5/bin/lupdate src -ts translations/settings-patchmanager.ts`


Olf0:
As I discovered the handful of untranslated `de` strings, I also reran `lupdate` in order to ensure that I did not forget even more: Luckily I did not, only some line number references were updated.  Still commit it, as I have now performed almost all steps required for that.

Issue #521, opened 2025-11-03
[Bug] PM operational, but enabled patches show no effect - after system update


nephros:
**SailfishOS VERSION**: 5.0.0.71 (Tampella) (observed on this, likely any)
<br />**HARDWARE**: Jolla C2 (likely any) 
<br />**PATCHMANAGER VERSION**: 3.2.12
<br />

#### BUG DESCRIPTION

All patchmanager GUI functions operate normally, patches can be installed, uninstalled, activated, de-activated and so on.

But none of the active patches show any effect on the system.

#### STEPS TO REPRODUCE

Details Unknown, but observed after a system update from 5.0.0.70 to 5.0.0.71, by myself, FSO user explit, and some others (Telegram channel).

#### ADDITIONAL INFORMATION

Re-installing PM fixes the issue.

nephros:
Now, thinking this through:

> But none of the active patches show any effect on the system.

This means there's some problem with the preload library. Either it is not loaded at all, or communication with the daemon does not work correctly.

> Re-installing PM fixes the issue.

Combined with the observation above, this hints to this section in rpm `%post`:

https://github.com/sailfishos-patches/patchmanager/blob/14677a3df89f4ce482be0dd78b5573648cfff395/rpm/patchmanager.spec#L194-L201

Theory being: a system update overwrites `/etc/firejail/whitelist-common.local` and/or `/etc/ld.so.preload`.  
That would cause the preload library not being in effect.

A re-install of the PM package obviously corrects this, and PM is again functioning normally.

Both of these files are owned by `droid-config-$DEVICEMODEL`, so updates of this package should cause the bug.

nephros:
A possible solution is to move the preload/filejail setup currently done in the .spec `%post/postun...` sections to a [oneshot script](https://github.com/sailfishos/oneshot), to be run at every boot.  
This could ensure the infrastructure is set up correctly at every boot.

@Olf0 do you think that is feasible?

Issue #39, opened 2019-03-19
Do not start pm in upgrade mode


CODeRUS:
Add os-update.target to conflicts of pm service

Olf0:
Thanks, this potentially closes issue #33, too.
Will test.

nephros:
I'm sorry, I'd like to get some context on this.

Change proposed is to add a `Conflicts=os-update.target ` entry to `/usr/lib/systemd/system/dbus-org.SfietKonstantin.patchmanager.service`, correct?

With the goal to not interfere with anything going on during a system update.

Qs:

1. is this still relevant in the days of PM3 and SFOS 4.x?
2. if yes, what creates os-update.target? Is it reliably* a thing that indicates a device is currently updating?
2. if not, what else might be used?

[*] with definitions of reliable adapted to the reality of how SFOS/Jolla changes happen... ;)

Olf0:
As:
1. Yes, I believe so: It does not make any sense for PM's backgroud services to start, when `os-update.target` is active.
2. a. SailfishOS!
    b. Well, it is reliably indicating that a SailfishOS installation has booted into the "single user upgrade phase" of an OS upgrade via GUI.
        You know (if not upgrading SFOS at the command line), one looks for an SFOS-upgrade in the Settings app, downloads its RPMs there; then the device reboots to `os-update.target`, only a progress bar is shown while the RPMs are installed and finally the device switches (or reboots, I do not remember precisely) to the regular multiuser-with-UI target.
       Mind that this method of upgrading SFOS cannot be used on community ports and "Sailfish Free / Trial" installations (i.e., SFOS installations without a paid license from Jolla) and there are may other ways of upgrading SFOS (the "zypper dance" / per `zypper --dup`, per `version dup`, per `pkcon upgrade-system`, which all not boot to a special target.
     TL;DR: `os-update.target` is a kind of "single user mode for upgrading SFOS", used when triggering an OS upgrade per Settings app.
  c. Well, this mechanism exists at least since SFOS 1.1.x (when I joined the club). 
3. Good question.  I guess the answer is "nothing", because the upgrading per command line (whichever way) cannot be detected, as it is just downloading and installing RPMs.
[...]

PR #523, opened 2025-12-15
Move preload and firejail setup to oneshot script

nephros:
Workarounds/fixes for #521

Instead of setting up files in `%post`, do it in a script and launch that
through the oneshot feature

### DIFF
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -182,7 +182,12 @@ exit 0
 
 %post
 export NO_PM_PRELOAD=1
-%{_bindir}/add-oneshot patchmanager-setup-preload.sh
+
+# set up the oneshot script, and run it immediately
+# If --now is given, job is run immediately instead of postponing it later
+# If instant run fails the link is created for later run
+%{_bindir}/add-oneshot --now patchmanager-setup-preload.sh
+
 case "$1" in
 1)  # Installation
   echo "Installing %{name}: %%post section"
--- a/rpm/patchmanager.spec
+++ b/rpm/patchmanager.spec
@@ -39,6 +39,7 @@ Requires:   grep
 Requires:   sed
 Requires:   sailfish-version >= 3.4.0
 Requires:   qml(Nemo.Configuration)
+Requires:   oneshot
 BuildRequires:  pkgconfig(Qt5Core)
 BuildRequires:  pkgconfig(Qt5DBus)
 BuildRequires:  pkgconfig(Qt5Qml)
@@ -53,6 +54,7 @@ BuildRequires:  pkgconfig(libshadowutils)
 BuildRequires:  qt5-qttools-linguist
 BuildRequires:  pkgconfig(rpm)
 BuildRequires:  pkgconfig(popt)
+BuildRequires:  oneshot
 
 
 %package testcases
@@ -180,6 +182,7 @@ exit 0
 
 %post
 export NO_PM_PRELOAD=1
+%{_bindir}/add-oneshot patchmanager-setup-preload.sh
 case "$1" in
 1)  # Installation
   echo "Installing %{name}: %%post section"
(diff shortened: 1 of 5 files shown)


b100dian (approved):
I like this solution, even though it is sailfishos specific (or because it is?), If @Olf0 does not have anything against I think we should merge and see if there are any more reports.

Olf0:
> If @Olf0 does not have anything against I think we should merge and see if there are any more reports.

Thanks; I had little time the past two months, but will now take a look at it soon™.

Olf0:
@nephros, can you please check my understanding denoted in commit ce0d96c, specifically if I expanded "the link" correctly.

I plan to thoroughly review the shell code proper soon™.

nephros:
> @nephros, can you please check my understanding denoted in commit [ce0d96c](https://github.com/sailfishos-patches/patchmanager/commit/ce0d96cd9b37640dd05d5b5abc71417a71eac17e), specifically if I expanded "the link" correctly.
> 
> I plan to thoroughly review the shell code proper soon™.

All fine! :) (the original comment was straight from the oneshot manual, but this version is better.

Now, what could happen is if this run of the oneshot "fails" for whatever reason, PM will not be functional (pretty much as described in #521).  
The situation should resolve itself after a reboot (provided the oneshot is executed then).

nephros:
I just did an update of my C2 to 5.0.0.73(EA), having installed a build of PM called `patchmanager-3.2.12+obs.20251217121522.165.g81845d3-1.17.1.bso.aarch64` which includes a variant of this PR, and I am still affected by #521.

`/etc/ld.so.preload` and `/etc/firejail/whitelist-common.local` are unchanged (from default) after the update, so it seems the oneshot script either never got executed, or failed changing the files.

I will have to investigate...

nephros:
> I just did an update of my C2 to 5.0.0.73(EA), having installed a build of PM called `patchmanager-3.2.12+obs.20251217121522.165.g81845d3-1.17.1.bso.aarch64` which includes a variant of this PR, and I am still affected by #521.
> 
> `/etc/ld.so.preload` and `/etc/firejail/whitelist-common.local` are unchanged (from default) after the update, so it seems the oneshot script either never got executed, or failed changing the files.
> 
> I will have to investigate...

 1. possible reason:

The preload macro is executed in the "at package install" section of the rpm `%post`. Therefore it is not executed when one updates from a PM package which used the old method.

Perhaps the fix for that is to call it also in the "at package update" case (but perhaps without the `--now` switch).

Changed in [629d735](https://github.com/sailfishos-patches/patchmanager/pull/523/commits/629d73559802ba291972ce8925f0fe8cc089570d).

[EDIT:] Tested that, and it works fine.

 2. possible reason:

If I understand correctly, the script is run exactly once (if a symlink exists in /etc/oneshot.d). If the run is successful, the symlink is removed.  
So if the timeline is

   - SFOS version X installed
   - PM installed -> oneshot script either executed at install, or at the next reboot.
   - After oneshot is run, symlink is removed
   - PM running OK
   - SFOS version updated from X->Y -> preload and whitelist files replaced by default ones
   - boot during/after update -> no oneshot script is run, default files remain in place
   - --> Bug #521 

For this issue, we could either "always fail" (`exit 1`) in the script.  
Or come up with something that is only done on system update. Some insights from the ancient [#39](https://github.com/sailfishos-patches/patchmanager/issues/39) may help here.

nephros:
> Or come up with something that is only done on system update. Some insights from the ancient [#39](https://github.com/sailfishos-patches/patchmanager/issues/39) may help here.

I wonder:

Make a service like this: 

```
[Unit]
...
After=system-update-pre.target
After=sailfish-upgrade-ui.service
Before=system-update.target

[Service]
Type=oneshot
RemainAfterExit=True

ExecStart=/bin/true
ExecStop=setup-patchmanager-foo
```


This should then run `setup-patchmanager-foo` once at the end of a system update, right?

If that is true, let `setup-patchmanager-foo` be `add-oneshot patchmanager-setup-preload.sh` and we're good?

nephros:
Soo, what is the chance that [8d4e507](https://github.com/sailfishos-patches/patchmanager/pull/523/commits/8d4e507a6baf6c94cd5e14fd52ddea832712e341) actually "bricks" devices by causing a failure during OS updates? Or regular boot?  

Do we risk it?

PR #514, opened 2025-07-02
Add new SVG icon, along with PNG exports

Olf0:
For its original source, see https://codeberg.org/marbalf/Sailfish-icons/issues/3

PR #525, opened 2025-12-17
Yet another Icon fix

nephros:
PR #514 renamed the icon, but did not change the entry in the Settings plugin config. This made PM appear icon-less once more in the Settings entries list.


nephros:
Note that while this does make the icon appear, PR #513 introduced another issue:

The icon is too wide, pushing the Patchmanager name to the right and thus not aligning with the rest of the Settings entries in the list.

b100dian (approved):

PR #527, opened 2025-12-17
Drop some legacy code

nephros:
The days of ausmt are long gone. I don't believe there is an upgrade path from
an ausmt-using system to current Patchmanager version.

Same thing with the `resetSystem` feature, I regard this as unmaintained and very much untested legacy code.

Wrap this code in some #ifdefs, we may drop it later at some point.


b100dian (approved):
👍 for these kind of small cleanups

Olf0:
> I don't believe there is an upgrade path from
> an ausmt-using system to current Patchmanager version.

Yes, IIRC one had to revert all Patches in PM2 to migrate to PM3.

PR #532, opened 2026-07-21
Daemon: do not start during OS update

nephros:
A bit hard to test.

The ConditionPathExists also appears this way in the lipstick service,
therefore I conclude it's the preferred way of handling this.


b100dian (approved):
Is safe to release though, even if it didn't work, it will not crash and burn.

PR #533, opened 2026-08-04
Bitness mangling: default to true

nephros:
This was set to false when mangling was a new feature.
It's been around for a while now and seems to work well.

Switch it on per default.


b100dian (approved):
Fully agreed :)

